PR #19, opened 2017-11-01
Update `Help with fees` step to match prototype

zheileman:
NOTE: the show/hide is not working yet.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -97,6 +97,8 @@ en:
           attributes:
             help_paying:
               inclusion: Select if you need help paying the court fee
+            hwf_reference_number:
+              blank: Please enter your reference number
 
   helpers:
     fieldset:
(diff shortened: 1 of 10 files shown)


config/locales/en.yml
```diff
@@ -111,7 +113,10 @@ en:
         number_of_children: Enter the number of children
       steps_help_with_fees_help_paying_form:
         help_paying:
-          <<: *YESNO
+          yes_with_ref_number: Yes, I have a Help with Fees reference number
+          yes_without_ref_number: Yes, I don’t have a Help with Fees reference number
```
digitalronin: Please change this to; `Yes, but I don’t have a Help with Fees reference number`
I think that's a lot clearer, in context. I have run this past Dave H.

zheileman: Ok, I was not aware of that. I will change it.

PR #24, opened 2017-11-06
Add Respondent task and steps with CRUD

zheileman:
This mimics the Applicant CRUD and allows to add 1 or more Respondant.
Right now there is no limitation on how many you can add but apparently
there will be a limit of 2 (to be clarified).

Also, there is (on purpose) duplication as we have now 2 very similar
form objects and views, so next step will be to DRY all this, including
i18n and specs (potential to extract shared examples).

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -131,9 +131,9 @@ en:
   helpers:
     fieldset:
       steps_applicant_personal_details_form:
-        has_previous_name: Do you have a previous name?
+        has_previous_name: Previous name?
       steps_respondent_personal_details_form:
-        has_previous_name: Do you have a previous name?
+        has_previous_name: Previous name?
         dob_unknown_html: ""
         postcode_unknown_html: ""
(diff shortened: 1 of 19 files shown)


config/locales/en.yml
```diff
@@ -103,10 +112,32 @@ en:
...
+            home_phone:
+              blank: Please enter the home phone
+            email:
+              invalid: Please enter a valid email
   helpers:
     fieldset:
       steps_applicant_personal_details_form:
         has_previous_name: Do you have a previous name?
+      steps_respondent_personal_details_form:
+        has_previous_name: Do you have a previous name?
```
digitalronin: The applicant is the one filling in the form, so the respondent is a different person. I think this needs to be "Do they have a previous name?"

zheileman: Yep, good catch. Actually I've checked the prototype and in both, applicant and respondent, it just says "Previous name?" so I will change it to match the prototype.

PR #30, opened 2017-11-08
Add new children additional details step

zheileman:
This form view has quite a lot of show/hide elements controlled by radio options. As we don't have a functional show/hide mechanism at the moment, all fields will be visible at once in the view, but the validations behind are implemented based on the prototype rules.

Talking about validations, for these kind of form objects are quite verbose and repetitive. Might be a potential area to try to refactor or come with a more cleaner way of doing it in the future.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -86,6 +86,10 @@ en:
           page_title: Child personal details
           heading: Child details
           btn_add_another: Add another child
+      additional_details:
+        edit:
+          page_title: Further information about the child(ren)
+          heading: Further information about the child(ren)
 
   home:
     index:
@@ -161,6 +165,8 @@ en:
         steps/children/personal_details_form:
           attributes:
             <<: *PERSONAL_DETAILS_ERRORS
+        steps/children/additional_details_form:
+          attributes:
 
   helpers:
     fieldset:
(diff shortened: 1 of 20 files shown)


config/locales/en.yml
```diff
@@ -196,6 +207,22 @@ en:
...
+          <<: *YESNO
+        children_protection_plan:
+          <<: *YESNO
+        children_same_parents:
+          <<: *YESNO
+        children_residence:
+          applicant: Applicant(s)
+          respondent: Respondent(s)
+          other: Other
+        children_known_to_authorities_details: Please state which child and the name of the Local Authority and Social worker (if known)
```
digitalronin: Should this be `which child/children` ?

config/locales/en.yml
```diff
@@ -196,6 +207,22 @@ en:
...
+        children_same_parents_yes_details: What are the names of the parents?
+        children_same_parents_no_details: Please give details of each parent and their children involved in this application
+        children_parental_responsibility_details: Please state everyone who has parental responsibility for each child and how they have parental responsibility
+        children_residence_details: Please give the full address of the child, the names of any adults living with the children and their relationship to or involvement with the child
```
digitalronin: Same here - child/children ... relationship(s) ... the child/children

PR #53, opened 2017-11-23
Add MIAM attended and certification steps

zheileman:
More MIAM questions, and updated decision trees.

Some of these step views are placeholders awaiting final copy and might change, or be removed.

### DIFF
--- a/app/views/steps/miam/no_certification_kickout/show.en.html.erb
+++ b/app/views/steps/miam/no_certification_kickout/show.en.html.erb
@@ -1,0 +1,21 @@
+<% title 'You need a MIAM certificate' %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= step_header %>
+
+    <!-- This is a temporal placeholder -->
+
+    <div class="grid_content_header">
+      <h1 class="heading-xlarge gv-u-heading-xxlarge">You need a signed MIAM certificate from your mediator</h1>
+    </div>
+
+    <div class="govuk-govspeak gv-s-prose">
+      <ol class="list list-number">
+        <li>Go to mediator</li>
+        <li>Ask for certificate</li>
+        <li>Return to continue</li>
+      </ol>
+    </div>
+  </div>
+</div>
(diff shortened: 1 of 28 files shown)


app/views/steps/miam/no_certification_kickout/show.en.html.erb
```diff
@@ -0,0 +1,21 @@
+<% title 'You need a MIAM certificate' %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= step_header %>
+
+    <!-- This is a temporal placeholder -->
```
tatyree: You mean 'temporary'?  It that why the copy isn't in the translations?

app/views/steps/miam/not_attended_info/show.en.html.erb
```diff
@@ -0,0 +1,24 @@
+<% title 'MIAM not attended' %>
+
+<div class="grid-row">
+  <div class="column-two-thirds">
+    <%= step_header %>
+
+    <!-- This is a temporal placeholder -->
```
tatyree: You mean 'temporary'? It that why the copy isn't in the translations?

app/views/steps/miam/no_certification_kickout/show.en.html.erb
zheileman: Yep, temporary 😄

The fact it is not in translations is in effect because we don't know yet the final copy but also because for some static copy-heavy steps, without form objects, fields, etc. for this project we are going with translated files instead i.e. `show.en.html.erb` for english, etc. as that makes it easier to update/maintain than having a lot of keys in i18n or a huge key with embedded html.

tatyree: That's fine.  Can you change 'temporal' to 'temporary', please?

app/views/steps/miam/not_attended_info/show.en.html.erb
tatyree: Can you change 'temporal' to 'temporary', please?

PR #54, opened 2017-11-24
Add `asking orders` step

zheileman:
This step let the applicant choose what the court needs to do about
the children. When `other` is selected, it goes to another step to
enter more details.

### DIFF
--- a/lib/extensions/string_extension.rb
+++ b/lib/extensions/string_extension.rb
@@ -1,0 +1,15 @@
+module StringExtension
+  def true?
+    /(true|t|yes|y|1)$/i.match(to_s).present?
+  end
+
+  def false?
+    to_s.strip.empty? || /(false|f|no|n|0)$/i.match(to_s).present?
+  end
+
+  def to_bool
+    return true  if true?
+    return false if false?
+    raise ArgumentError, "Invalid value for Boolean: `#{self}`"
+  end
+end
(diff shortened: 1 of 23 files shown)


lib/extensions/string_extension.rb
```diff
@@ -0,0 +1,15 @@
+module StringExtension
+  def true?
+    /(true|t|yes|y|1)$/i.match(to_s).present?
+  end
+
+  def false?
+    to_s.strip.empty? || /(false|f|no|n|0)$/i.match(to_s).present?
+  end
+
+  def to_bool
```
tatyree: This isn't being used anywhere.  Do you need it at this point?

zheileman: Hmm, not really. I guess I can just leave the true? / false? methods. I will remove it and update the tests as at the moment to_bool is the method being tested than implicitly uses the others.

lib/extensions/string_extension.rb
```diff
@@ -0,0 +1,9 @@
+module StringExtension
+  def true?
+    /(true|t|yes|y|1)$/i.match(to_s).present?
+  end
+
+  def false?
```
tatyree: While it makes sense for logical completeness, I've just realised `.false?` isn't being used anywhere except the tests, either...

I'd suggest it should come out as well–makes the tests simpler and keeps potential mutants at bay 😄

zheileman: Ok. I would normally leave it there for the logical sense as you say, but probably in this service will never be used, as looking at the decision trees we do things when something is checked, not when something is *not* checked. Or we cover these under the If (checked) else (not checked) statement. So will remove it 👍

PR #116, opened 2018-01-22
Finish the safety MIAM exemptions

zheileman:
A bit of code refactor and extracting the exemptions to a value-object to make life easier.
More details in the individual commits.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -864,6 +864,7 @@ en:
       steps_miam_certification_number_form:
         miam_certification_number: FMC registration number
       steps_miam_exemptions_safety_form:
+        ## GROUP POLICE ##
         group_police_html: |
           <span class="heading-small gv-u-heading-small">The police have been involved</span><br/>
           <span class="form-hint">For example, you or the other person (the respondent) has been arrested or cautioned for domestic or child abuse offences.</span>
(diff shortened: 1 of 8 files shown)


config/locales/en.yml
```diff
@@ -872,6 +873,83 @@ en:
         police_ongoing_proceedings: Evidence of relevant criminal proceedings for a domestic violence offence which have not concluded
         police_conviction: Evidence of a relevant conviction for a domestic violence offence
         police_dvpn: A domestic violence protection notice issued under section 24 of the Crime and Security Act 2010 against a prospective party
+        ## GROUP COURT ##
```
tatyree: Just a suggestion: break the locales out categorically:

```
locales/en/group_police.yml
locales/en/police.yml
```
I find them easier to maintain this way, personally.

zheileman: Yeah I know. In tax tribs we originally had multiple i18n files, but then we decided to unify all of them into a big monolith because for shared keys it simplified the work.

But again, tax tribs was a baby compared to this project, so yes, I'm every other day trying to decide if we should split the i18n or not and how to best organise it (I'm not a big fan of many small i18n but a few is ok, for example one for exemptions seems quite logical).

tatyree: Maybe it's worth looking at one of the Rails `gettext` implementations at this point?

PR #130, opened 2018-01-25
Protect from a nil asking order

zheileman:
So we can render and iterate the PDF quickly, without having to go
through all of the questions. In this case, the render was blowing
up if we didn't answer the `Petition` orders step.

### DIFF
--- a/spec/presenters/petition_presenter_spec.rb
+++ b/spec/presenters/petition_presenter_spec.rb
@@ -66,4 +66,17 @@ RSpec.describe PetitionPresenter do
       )
     end
   end
+
+  describe '#other_details' do
+    context 'when there is an asking order' do
+      it { expect(subject.other_details).to eq('details') }
+    end
+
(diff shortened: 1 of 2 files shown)


spec/presenters/petition_presenter_spec.rb
```diff
@@ -66,4 +66,15 @@
       )
     end
   end
+
+  describe '#other_details' do
+    context 'when there is an asking order' do
+      it { expect(subject.other_details).to eq('details') }
+    end
+
+    context 'when the asking order is nil' do
```
digitalronin: The important point here is that the presenter doesn't fail, isn't it? So, shouldn't this spec say explicitly that an error/exception doesn't happen?

zheileman: Yep, I agree. I've added an explicit expectation to check it doesn't raise the exception.

PR #132, opened 2018-01-25
Add court postcode screener

aldavidson:
Add a simple screening question as the first 'screener' -  all suggestions on copy wording, design of the form, and workflow are very welcome! 

The entered postcodes are then checked against the Courtfinder API for the 'Children' area of law, and matched against a list of courts taking part in this service. If at least one is found, the user can continue. If not, they are shown a 'sorry, you are ineligible' page (again, please suggest improvements for the copy)


app/forms/steps/postcode_screen/children_postcodes_form.rb
```diff
@@ -0,0 +1,19 @@
+module Steps
+  module PostcodeScreen
+    class ChildrenPostcodesForm < BaseForm
+      include SingleQuestionForm
```
zheileman: You can remove this include, as the `SingleQuestionForm` is only intended for steps with a Yes/No question, but as this step is a text area, it doesn't apply really.
And once you remove it, instead of using `record_to_persist`, you have to use `c100_application`
Take this other form object as an example as it should be almost identical: https://github.com/ministryofjustice/c100-application/blob/master/app/forms/steps/application/details_form.rb

app/views/steps/postcode_screen/error_but_continue/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <%= step_header %>
+
+    <h1 class="heading-xlarge gv-u-heading-xxlarge"><%=t '.heading' %></h1>
+
+    <p class="lede"><%=t '.lead_text' %></p>
+
+    <!-- TODO: Add the copy needed, and adapt the button linking to another step -->
+
+    <div class="xform-group form-submit">
+      <%= link_to t('.go_back'), edit_steps_postcode_screen_children_postcodes_path, class: 'link-back link-button' %>
```
zheileman: Wouldn't be better to just use the 'back' link we have in the header in all the steps? `<%= step_header %>`

app/services/courtfinder_api.rb
```diff
@@ -0,0 +1,43 @@
+require 'json'
+require 'open-uri'
+
+class CourtfinderAPI
```
zheileman: Can you put this service inside the namespace `C100App`? That way, mutant will get to test it automatically (anything inside the `C100App` namespace gets mutant-tested).

app/services/court_postcode_checker.rb
```diff
@@ -0,0 +1,33 @@
+require 'courtfinder_api'
+
+class CourtPostcodeChecker
```
zheileman: Can you put this service inside the namespace C100App? That way, mutant will get to test it automatically (anything inside the C100App namespace gets mutant-tested).

app/forms/steps/postcode_screen/children_postcodes_form.rb
aldavidson: done

app/controllers/steps/postcode_screen_step_controller.rb
```diff
@@ -0,0 +1,9 @@
+module Steps
+  class PostcodeScreenStepController < StepController
+    private
+
+    def decision_tree_class
+      C100App::PostcodeScreenDecisionTree
```
zheileman: I've realised that what we probably want is a "Screener" task, with many steps, as the screener will be composed of multiple steps, not only the postcode step. That way we have a single decision tree handling all the screener logic.

digitalronin: That's true, but we do also need a separate postcode step for near the end of the process, when we tell the user where they need to send their printed application.

zheileman: So the way I understand it, is a Task screener, composed of N-steps, one of them being the postcode 'whitelisting', and if the postcode passes our whitelist/rules for the pilot, we have it already stored in DB and by the end of the process (granting the rest of the screening questions were successful), based on this postcode, we show the correct instructions. But there is no separate step for postcode at the end, is just the 'Completion' step, with instructions.

app/services/court_postcode_checker.rb
aldavidson: done

app/services/courtfinder_api.rb
aldavidson: done

app/views/steps/postcode_screen/error_but_continue/show.html.erb
aldavidson: sure -  I went for the second explicit link as the copy is implying a choice between two possible actions, so presenting the options next to each other seemed to make sense. I've taken it out now though.

app/controllers/steps/postcode_screen_step_controller.rb
aldavidson: my thinking was that in the Completion step, you'd have something like:

What to do now
===========
Print the form, sign it, and post it to:

<the appropriate court address>

(button: Print the form)

zheileman:
Hi Al. I'm approving the PR as it is only pending some mutant tests failures 👍  Once that's solved you can go ahead and merge it. Remember you can run mutant on individual files, for example:
`bundle exec rake mutant Steps::PostcodeScreen::ChildrenPostcodesForm`
or you can run the whole project with `bundle exec rake mutant all`

PR #151, opened 2018-02-01
 Add `other parties` details to PDF

zheileman:
Including a bit of refactor of the relationships to a presenter to be reused in several of the sections.

screen shot 2018-02-01 at 12 52 15

### DIFF
--- a/app/presenters/relationships_presenter.rb
+++ b/app/presenters/relationships_presenter.rb
@@ -1,0 +1,35 @@
+class RelationshipsPresenter
+  RELATION_SEPARATOR = ' - '.freeze
+  PEOPLE_SEPARATOR = '; '.freeze
+
+  attr_reader :c100_application
+
+  def initialize(c100_application)
+    @c100_application = c100_application
+  end
+
+  def relationship_to_children(person_or_people, show_person_name: true)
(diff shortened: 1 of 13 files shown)


app/presenters/relationships_presenter.rb
```diff
@@ -0,0 +1,35 @@
+class RelationshipsPresenter
```
digitalronin: Is it worth having a base presenter class, maybe inheriting from Delegator to forward attribute calls to the c100_application object?
https://ruby-doc.org/stdlib-2.4.0/libdoc/delegate/rdoc/Delegator.html

zheileman: Hmm, I don't think in this particular case would be worth it, as we are only accessing 2 methods from `c100_application` (#children and #relationships). But if we start creating other small presenters like this one, accessing more c100 methods/attributes, then it could be a good idea, yes.

digitalronin: Yeah, that's what I meant. I think we have a couple of presenters that mostly delegate calls to the c100app object. If there's a pattern there, extracting a superclass might be sensible, and inheriting from delegator could simplify the code.

PR #155, opened 2018-02-05
Implement `Other children` relationships

zheileman:
This was missing from the rails app (and I think also the prototype) but was confirmed with Andy it is needed.

Due to this oversight we didn't model properly the Child and OtherChild in first place and now a substantial refactor had to be made, introducing an intermediate model called `Minor` (inheriting from `Person`) and then our current models `Child` and `OtherChild` can inherit from `Minor`.
This allows us to create relationships between a 'person' (being these applicant, respondent or other party) and a 'minor' (being these child or other child). And to retrieve all kind of children using `#minors`.

screen shot 2018-02-05 at 11 45 53

### DIFF
--- a/app/models/relationship.rb
+++ b/app/models/relationship.rb
@@ -1,6 +1,6 @@
 class Relationship < ApplicationRecord
   belongs_to :c100_application
 
-  belongs_to :child,  foreign_key: :child_id
-  belongs_to :person, foreign_key: :person_id
+  belongs_to :minor
+  belongs_to :person
 end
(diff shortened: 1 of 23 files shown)


app/models/relationship.rb
```diff
@@ -1,9 +1,6 @@
 class Relationship < ApplicationRecord
   belongs_to :c100_application
 
-  belongs_to :child,       foreign_key: :child_id
-  belongs_to :person,      foreign_key: :person_id
-  belongs_to :applicant,   foreign_key: :person_id
-  belongs_to :respondent,  foreign_key: :person_id
-  belongs_to :other_party, foreign_key: :person_id
```
aldavidson: 👍 this looks much cleaner

PR #158, opened 2018-02-05
Refactor postcode screen to generic screener

aldavidson:
There should be no visible difference to the user, but behind the scenes:
* screener answers are now stored in a dependent record (screener_answers)
* the functionality that was in the PostcodeScreenXYZ classes has all been refactored into a generic Screener task, with a Postcode step


db/migrate/20180205113216_add_screener_answers.rb
```diff
@@ -0,0 +1,9 @@
+class AddScreenerAnswers < ActiveRecord::Migration[5.1]
+  def change
+    create_table :screener_answers do |t|
```
zheileman: We use `uuids` for primary keys in all tables. Not that it really matters in this table as we are not going to expose the records publicly for CRUD or anything like that, but for consistency, it might be a good idea to continue using uuids. Just adding `id: :uuid` will do the trick. Otherwise it generates a numeric ID pk by default.

Also, maybe timestamp fields (`t.timestamps null: false`) could be interesting if we were to do internal analysis of records in this table, to order and group by date, and create reports 🤔

db/migrate/20180205113239_move_children_postcodes_to_screener_answers.rb
```diff
@@ -0,0 +1,15 @@
+class MoveChildrenPostcodesToScreenerAnswers < ActiveRecord::Migration[5.1]
+  def up
+    C100Application.all.each do |app|
+      app.screener_answers.children_postcodes = app.children_postcodes
```
zheileman: This fails as, in my case, I have C100Application records but those are not yet linked to any `screener_answer` record so it returns `nil` and 💥 
I would personally get rid of this migration altogether as we've not released the service publicly yet and there is no harm in loosing local/test data. It might come back to bite us later.

db/schema.rb
```diff
@@ -290,4 +296,6 @@
   add_foreign_key "relationships", "c100_applications"
   add_foreign_key "relationships", "people"
   add_foreign_key "relationships", "people", column: "minor_id"
+  add_foreign_key "relationships", "people", column: "child_id"
```
zheileman: This line needs to go. The valid one is the other, with `minor_id`

spec/controllers/steps/postcode_screen/children_postcodes_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
-require 'rails_helper'
```
zheileman: To be removed.

spec/controllers/steps/postcode_screen/children_postcodes_controller_spec.rb
```diff
@@ -1,6 +1,6 @@
...
+# require 'rails_helper'
```
zheileman: To be removed.

spec/forms/steps/postcode_screen/children_postcodes_form_spec.rb
```diff
@@ -1,44 +1,44 @@
-require 'spec_helper'
```
zheileman: To be removed.

aldavidson:
rebased from master & pushed

PR #162, opened 2018-02-07
Implement C8 details confidentiality

zheileman:
If the applicant enables the confidentiality of personal details,
then the C100 PDF form will hide a limited set of details and will
generate an extra C8 form with the omitted details.

This PR only implements the 'hiding' of details in a flexible way,
but not yet the output of the C8 form.

screen shot 2018-02-07 at 12 20 10


zheileman:
⏳

PR #166, opened 2018-02-12
Implement `C8 form` people details

zheileman:
A refactor of the C8 presenter had to be done in order to simplify the output of personal details for applicants and other parties in the C8.

screen shot 2018-02-09 at 16 39 54


zheileman:
@tatyree yes, it is fragile but on purpose. I mean, in the dev phase it is really a chore to have to update all those specs just because we decided to introduce a new row, or remove another. But once it is finished and stable, it will raise a red flag if it ever gets out of sync, which means a quite solid safety net for maintainers (HMCTS?) in the future.

I guess with a bit of extra code we can make the array index a variable that auto-increments, so it is not hardcoded and we don't have to re-adjust the indexes when we introduce a new row in the middle. Although this extra cleverness may be at the expense of readability. I can give it a try to see if I come up with a not very ugly compromise 😄

tatyree:
I figured that it was something like that. No need to do anything special.

tatyree:
Actually... It might be worth putting an abridged version of your response
in a comment on the spec to keep tooth-suckers such as myself from
constantly going "you know this is fragile, right?"

zheileman:
@tatyree Yes, I can add some comments in the next PR for these "pdf row" tests.

PR #167, opened 2018-02-12
Add `admin` section to C100 PDF

zheileman:
A bit of refactor to reuse components for both the C8 and C100 sections.

Also, added a bit of explanation about fragile tests in those specs with a high number of rows, as these are more prone to be affected with code changes.

screen shot 2018-02-12 at 11 41 45


spec/presenters/summary/sections/other_children_details_spec.rb
```diff
@@ -34,6 +34,11 @@ module Summary
       end
     end
 
+    # The following tests can be fragile, but on purpose. During the development phase
+    # we have to update the tests each time we introduce a new row or remove another.
+    # But once it is finished and stable, it will raise a red flag if it ever gets out
+    # of sync, which means a quite solid safety net for any maintainers in the future.
+    #
```
tatyree: 👍

PR #210, opened 2018-02-19
add are-you-a-parent screener question & exit page

aldavidson:
This PR will insert a screener question for "Are you a parent?" in between the Urgency and MIAM Exemptions. The content should be considered placeholder only, and should be revised by a content designer.

screen shot 2018-02-19 at 16 49 01

If the user answers no:

screen shot 2018-02-19 at 16 49 08

### DIFF
--- a/app/controllers/steps/screener/parent_controller.rb
+++ b/app/controllers/steps/screener/parent_controller.rb
@@ -1,0 +1,15 @@
+module Steps
+  module Screener
+    class ParentController < Steps::ScreenerStepController
+      def edit
+        @form_object = ParentForm.new(
+          c100_application: current_c100_application
+        )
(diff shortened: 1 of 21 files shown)


app/controllers/steps/screener/parent_controller.rb
```diff
@@ -0,0 +1,15 @@
+module Steps
+  module Screener
+    class ParentController < Steps::ScreenerStepController
+      def edit
+        @form_object = ParentForm.new(
```
zheileman: You want to use `#build` here instead of `#new`, otherwise the value of the attribute is not retrieved for rendering.
In some forms we use #new but then we also need to pass the attribute(s) to set the values.

PR #237, opened 2018-02-26
add email contact screener question

aldavidson:
screen shot 2018-02-26 at 14 44 36

### DIFF
--- a/app/forms/steps/screener/email_consent_form.rb
+++ b/app/forms/steps/screener/email_consent_form.rb
@@ -5,9 +5,9 @@ module Steps
       include HasOneAssociationForm
 
       has_one_association   :screener_answers
-      yes_no_attribute      :email_consent
-      attribute             :email_address, String, reset_when_no: [:email_consent]
-      validates_presence_of :email_address, if: -> { email_consent&.yes? }
+      yes_no_attribute      :email_consent, reset_when_no: [:email_address]
+      attribute             :email_address, String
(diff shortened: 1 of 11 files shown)


app/forms/steps/screener/email_consent_form.rb
```diff
@@ -0,0 +1,24 @@
+module Steps
+  module Screener
+    class EmailConsentForm < BaseForm
+      include SingleQuestionForm
+      include HasOneAssociationForm
+
+      has_one_association   :screener_answers
+      yes_no_attribute      :email_consent
+      attribute             :email_address, String, reset_when_no: [:email_consent]
```
zheileman: The `reset_when_no` goes in the `yes_no_attribute` declaration. And we want to reset the email field, so it would be like this:

```ruby
yes_no_attribute :email_consent, reset_when_no: [:email_address]
attribute :email_address, NormalisedEmail
```

app/forms/steps/screener/email_consent_form.rb
```diff
@@ -0,0 +1,24 @@
...
+      validates_presence_of :email_address, if: -> { email_consent&.yes? }
```
zheileman: We have an email validator, so for this, we can do:

```ruby
validates :email_address, email: true, if: -> { email_consent&.yes? }
```

aldavidson: good call - done and pushed

PR #243, opened 2018-02-28
Refactor petition orders to use value-object

zheileman:
This follows the same mechanism as the MIAM exemptions, so we don’t
need to create dozens of DB fields, just we use a value-object to hold
the different values and can be grouped and extended/modified more
easily.

Lots of copy changes.

screen shot 2018-03-01 at 10 15 25

### DIFF
--- a/app/value_objects/petition_order.rb
+++ b/app/value_objects/petition_order.rb
@@ -4,7 +4,6 @@ class PetitionOrder < ValueObject
       new(:child_arrangements_home),
       new(:child_arrangements_time),
       new(:child_arrangements_contact),
-      new(:child_arrangements_event),
       new(:child_arrangements_access),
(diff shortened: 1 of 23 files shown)


app/value_objects/petition_order.rb
```diff
@@ -1,28 +1,39 @@
...
+      new(:specific_issues_medical),
+      new(:specific_issues_moving),
+      new(:specific_issues_moving_abroad),
     ].freeze,
 
-    OTHER = [
-      new(:other)
-    ].freeze
+    # This is the exception to the rule. The values here (one for now),
+    # don't have a corresponding order name. Because reasons.
```
tatyree: Kill the 'Because reasons.', please.  It's enough to know that they don't have an order name.

PR #283, opened 2018-03-15
Sort out expires

zheileman:
A few changes to make this consistent through the app.

Applications get purged after 28 days, and user accounts after 30 days since last time they signed-in.


config/application.rb
```diff
@@ -49,8 +49,8 @@ class Application < Rails::Application
   # and if it has an owner (it was saved for later), we send up to two email reminders
   # before we delete the application. If you change this number make sure to also update
   # `app/services/c100_app/reminder_rule_set.rb` and the Notify email templates.
-  config.x.drafts.expire_in_days = ENV.fetch('DRAFTS_EXPIRE_AFTER', 14).to_i
```
aldavidson: was there any particular reason to move away from an ENV variable with default?

zheileman: Yep, I explained it in the commit message:

"Do not using ENV variables for these expires, because there is some code, like the reminders, that rely on specific dates to trigger, and also tests that would fail.

This can be improved of course but I don't think is worth it as these numbers are not suppose to be changing every week, once we pick ones we will stick to them."

aldavidson: Seems like the best approach would be to remove that reliance on specific dates?
Not a blocker - I'll still approve this PR - but something to look at again in future

zheileman: Yep, no problem. We can look at this later 👍 
I just think we tend to bloat applications with ENV vars. Tax Tribs is horrible in this sense, full of env vars that almost never are going to change. Some of these were propagated from Tax Tribs to C100.

But happy to look at this and come up with a better approach. ENV vars are problematic also in the sense that if they get cached, like in this case, at boot time, then no matter if you change them in Heroku, you still can't get the app to use the new value, unless triggering a restart or deploy.

If env vars are prone to be hot-changed then these can't be cached at boot time (so using `application.rb` is not an option). We might want to do an audit of all the ENV vars currently in use, to see which ones are really needed, and which ones are more "constants" or even configuration to put in a yaml file, for example.

PR #291, opened 2018-03-19
add postcode format validation to screener form

aldavidson:
screen shot 2018-03-19 at 15 01 27


app/forms/steps/screener/postcode_form.rb
```diff
@@ -1,3 +1,18 @@
+class FullUkPostcodeValidator < ActiveModel::EachValidator
```
zheileman: As this is our first custom validator, can we move this class to `app/validators`? Rails will autoload it.

Also, do we also want to validate blank postcodes? at the moment it is letting the user pass (to the kickout page) instead of triggering a `:blank` error, not sure if that's on purpose.

aldavidson: that was a decision we took when we first discussed the screener - OTOH I agree it doesn't feel right now. I'll change that to require presence, and move it to app/validators

PR #299, opened 2018-03-21
Make well-formed but non-existent postcodes a validation error

aldavidson:
screen shot 2018-03-21 at 11 53 55


app/validators/full_uk_postcode_validator.rb
```diff
@@ -1,14 +1,37 @@
 class FullUkPostcodeValidator < ActiveModel::EachValidator
+  MAPIT_URI_ROOT = "https://mapit.mysociety.org/postcodes/".freeze
```
zheileman: I think the URL is `/postcode/` singular, otherwise it always returns a 404.

aldavidson: you're right! I think that's an error that crept in during the Mutant-wrangling - fixed now

aldavidson:
superceded by #303

PR #339, opened 2018-04-03
Add missing 404 page

zheileman:
This will show a "nice" 404 page when hitting an unrecognized URL.

On localhost however it will continue to show the error trace to help debugging.

screen shot 2018-04-03 at 11 45 04

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -964,6 +964,15 @@ en:
       lead_text: Your application won’t be processed until you complete a few more steps. If you want to make changes you can start a new application.
       what_next: What you need to do next
       <<: *START_FINISH
+    not_found:
+      page_title: Page not found
+      heading: Page not found
+      lead_text: If you copied a web address, please check it’s correct.
+      paper_form_html: |
+        <p>You can also try starting again from the homepage.</p>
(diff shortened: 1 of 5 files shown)


config/locales/en.yml
```diff
@@ -964,6 +964,15 @@ en:
       lead_text: Your application won’t be processed until you complete a few more steps. If you want to make changes you can start a new application.
       what_next: What you need to do next
       <<: *START_FINISH
+    not_found:
+      page_title: Page not found
+      heading: Page not found
+      lead_text: If you copied a web address, please check it’s correct.
+      paper_form_html: |
+        <p>You can also try starting again from the homepage.</p>
+        <p>If the problem continues, you can <a href="https://formfinder.hmctsformfinder.justice.gov.uk/c100-eng.pdf" target="external_link" rel="external">download and
```
solidgoldpig: Is this the canonical address for the form? I guess it is, since it redirects to the new GOV.UK page - but just checking...

zheileman: Yep that seems to be the URL. We are using it in other places in the app. As part of a little future refactor we might want to extract it to declare the URL in one single place for easy change if needed.

solidgoldpig: Cool

PR #341, opened 2018-04-03
reset the MIAM cert details when the user changes 'attended' to 'no'

aldavidson:


spec/forms/steps/miam/attended_form_spec.rb
```diff
@@ -2,4 +2,29 @@
 
 RSpec.describe Steps::Miam::AttendedForm do
   it_behaves_like 'a yes-no question form', attribute_name: :miam_attended
```
zheileman: I think the macro for the yes-no spec should give you for free the reset tests (but only if the attributes belong to the same table being updated), if you use: `reset_when_no: [:attrib1, attrib2...]`. Check `/spec/forms/steps/application/litigation_capacity_form_spec.rb` for an example.

aldavidson: ah-ha, you're right. I'll change it to use that method.

PR #346, opened 2018-04-05
Add sentry user-feedback to the unhandled error page

zheileman:
This nice feature from Sentry will show a modal window for users to provide optional feedback about what caused the error, including providing their email for a potential follow-up.

The modal can be closed without providing feedback, as it is optional, and also it will not show without JS (obviously) but will not stop our normal error page from showing.

screen shot 2018-04-04 at 17 04 15

### DIFF
--- a/app/assets/stylesheets/local/app.scss
+++ b/app/assets/stylesheets/local/app.scss
@@ -71,5 +71,11 @@
   margin: -1rem 0 2rem;
 }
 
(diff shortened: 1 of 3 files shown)


zheileman:
This is how it looks in Sentry. The feedback, if provided by the users, is attached to the exception:

screen shot 2018-04-05 at 10 37 01

PR #372, opened 2018-04-24
Add alternatives section to "check your answers"

aldavidson:
screen shot 2018-04-24 at 16 10 22


app/presenters/summary/html_sections/alternatives.rb
```diff
@@ -0,0 +1,37 @@
...
+                     change_path: edit_steps_alternatives_collaborative_law_path),
+
+        ].select(&:show?)
+      end
+      # rubocop:enable Metrics/MethodLength
+
+      private
+
+      def petition
+        @_petition ||= PetitionPresenter.new(c100)
```
zheileman: We can get rid of this.

PR #374, opened 2018-04-25
Add check answers children details

aldavidson:
Also fix broken HTML structure in the summary, replacing wrapper DL's with wrapper divs, and pushing the DL elements to the individual answers.

_NOTE_ this does not include the "do you have any other children?" sections, that will be a separate PR

screen shot 2018-04-25 at 16 29 30

### DIFF
--- a/app/presenters/summary/html_sections/base_children_details.rb
+++ b/app/presenters/summary/html_sections/base_children_details.rb
@@ -3,26 +3,62 @@ module Summary
     class BaseChildrenDetails < Sections::BaseSectionPresenter
       protected
 
+      # rubocop:disable Metrics/MethodLength
       def personal_details(child)
         [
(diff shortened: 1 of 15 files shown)


app/presenters/summary/html_sections/base_children_details.rb
```diff
@@ -0,0 +1,65 @@
...
+          Answer.new(
+            :child_sex,
+            child.gender,
+            change_path: edit_child_details_path(child, 'gender_female')
+          ),
+        ]
+      end
+      # rubocop:enable Metrics/MethodLength
+
+      def relationships(child)
```
zheileman: I believe the relationships will be best placed in each of the applicant/respondent sections, when we have them, but for now it is ok to leave them here. We can move them around later on.

PR #388, opened 2018-05-02
Add check answers "Other children" details

aldavidson:
screen shot 2018-05-02 at 11 10 20


app/presenters/summary/html_sections/base_children_details.rb
```diff
@@ -32,24 +32,61 @@ def personal_details(child)
...
       private
 
       def edit_relation_path(child, people)
         format("/steps/applicant/relationship/%<id>s/child/%<child_id>s",
-               id: relationship(child, people).pluck(:id),
+               id: relationship(child, people).pluck(:person_id).first,
                child_id: child.id)
       end
 
+      def child_names_method(child)
```
zheileman: Not saying to do do it, but maybe it would better to have all these route methods in subclasses, (so we would have a children subclass and an other_children subclass) so the subclasses are in charge of knowing the routes, and then in the superclass just call a generic method, lie 'edit_names_path', 'edit_personal_details_path' and so on...

These if-else are going to be messy for applicants, respondents, other parties (and solicitors, when we introduce them). Better to encapsulate anything that is particular to each kind of 'Person' in their subclasses, and use a template-pattern.

config/routes.rb
```diff
@@ -199,6 +199,7 @@ def edit_routes(path)
       crud_step :personal_details, only: [:edit, :update]
       crud_step :contact_details,  only: [:edit, :update]
       edit_step :relationship, only: [] do
+        edit_routes ':id'
```
zheileman: Are we using this? If not, let's remove it.

spec/presenters/summary/html_sections/children_details_spec.rb
```diff
@@ -79,7 +93,7 @@ module Summary
         expect(answers[5]).to be_an_instance_of(MultiAnswer)
         expect(answers[5].question).to eq(:child_respondents_relationship)
         expect(answers[5].value).to eq(['mother'])
-        expect(c100_application).to have_received(:respondents)
+        expect(c100_application).to have_received(:respondents).at_least(:once)
```
zheileman: Can we also test the change_paths? Specially here is important as there are many subtleties with paths.
Take a look at https://github.com/ministryofjustice/c100-application/blob/master/spec/presenters/summary/html_sections/abduction_spec.rb#L54 for example.

app/models/minor.rb
```diff
@@ -1,3 +1,10 @@
 class Minor < Person
   has_many :relationships, source: :minor, dependent: :destroy
 end
+
+# hackhackhack to make sure that c100_application.minors
+# includes the subtypes in development
+# (autoloading only includes _KNOWN_ subclasses -
+# see http://guides.rubyonrails.org/autoloading_and_reloading_constants.html#autoloading-and-sti )
+require_dependency 'child'
```
zheileman: Good catch 👍

app/presenters/summary/html_sections/base_children_details.rb
aldavidson: you're right, I've refactored it and it's much cleaner now :+1:

aldavidson:
@zheileman I've refactored as you suggested, and added tests for change_path values on the answers. Ready for re-review

zheileman:
superb 👍

PR #391, opened 2018-05-02
Update `nature of application` section in CYA

zheileman:
We need to include in this section the protection orders question and details, if C1A triggered.

A bit of refactor to compact a bit this section so it doesn't take too much space and is simpler to read.

screen shot 2018-05-02 at 15 56 51


zheileman:
I'm changing this.

PR #400, opened 2018-05-08
add previous court cases to CYA

aldavidson:
screen shot 2018-05-08 at 11 57 31

### DIFF
--- a/app/presenters/summary/html_sections/other_court_cases.rb
+++ b/app/presenters/summary/html_sections/other_court_cases.rb
@@ -1,0 +1,36 @@
+module Summary
+  module HtmlSections
+    class OtherCourtCases < Sections::BaseSectionPresenter
+      def name
+        :other_court_cases
+      end
+
+      # rubocop:disable Metrics/AbcSize
+      def answers
+        return [Separator.not_applicable] unless previous_proceedings?
+
+        [
+          Answer.new(:has_other_court_cases, c100.children_previous_proceedings, change_path: edit_steps_application_previous_proceedings_path),
+          FreeTextAnswer.new(:court_proceeding_children_names,   court_proceeding.children_names, change_path: edit_steps_application_court_proceedings_path(anchor: 'steps_application_court_proceedings_form_children_names')),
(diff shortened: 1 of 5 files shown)


app/presenters/summary/html_sections/other_court_cases.rb
```diff
@@ -0,0 +1,36 @@
...
+        :other_court_cases
+      end
+
+      # rubocop:disable Metrics/AbcSize
+      def answers
+        return [Separator.not_applicable] unless previous_proceedings?
+
+        [
+          Answer.new(:has_other_court_cases, c100.children_previous_proceedings, change_path: edit_steps_application_previous_proceedings_path),
+          FreeTextAnswer.new(:court_proceeding_children_names,   court_proceeding.children_names, change_path: edit_steps_application_court_proceedings_path(anchor: 'steps_application_court_proceedings_form_children_names')),
```
zheileman: I think this is a perfect question to use the `AnswersGroup`, as we've done in others.
All the change links are very repetitive, and go to the same page (and it saves us the trouble of using the anchors). 

So basically, the first question is the yes/no, that's fine, and the following is the AnswerGroup where we group together all the FreeTextAnswer under a single `change` link.

It also visually looks nicer.

app/presenters/summary/html_sections/other_court_cases.rb
```diff
@@ -0,0 +1,36 @@
+module Summary
+  module HtmlSections
+    class OtherCourtCases < Sections::BaseSectionPresenter
+      def name
+        :other_court_cases
+      end
+
+      # rubocop:disable Metrics/AbcSize
+      def answers
+        return [Separator.not_applicable] unless previous_proceedings?
```
zheileman: Also we can get rid of this, as this is more for the PDF were we need to say something even if the question doesn't apply to the applicant, but in the CYA we just don't render anything.

app/presenters/summary/html_sections/other_court_cases.rb
aldavidson: done - I've updated the screenshot on the PR to reflect this

app/presenters/summary/html_sections/other_court_cases.rb
aldavidson: done

aldavidson:
@zheileman it now uses AnswersGroup as you suggested

PR #402, opened 2018-05-08
add without notice hearing details to CYA

aldavidson:
screen shot 2018-05-08 at 15 50 49

### DIFF
--- a/app/presenters/summary/html_sections/without_notice_details.rb
+++ b/app/presenters/summary/html_sections/without_notice_details.rb
@@ -15,15 +15,13 @@ module Summary
       private
 
       def without_notice_hearing_details
-        return [] unless c100.without_notice.eql?(GenericYesNo::YES.to_s)
-
         [
           FreeTextAnswer.new(:without_notice_details,            c100.without_notice_details),
           Answer.new(:without_notice_impossible,                 c100.without_notice_impossible),
           FreeTextAnswer.new(:without_notice_impossible_details, c100.without_notice_impossible_details),
           Answer.new(:without_notice_frustrate,                  c100.without_notice_frustrate),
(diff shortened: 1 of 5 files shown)


app/presenters/summary/html_sections/without_notice_details.rb
```diff
@@ -0,0 +1,30 @@
...
+        [
+          Answer.new(:without_notice_hearing, c100.without_notice, change_path: edit_steps_application_without_notice_path),
+          AnswersGroup.new(:without_notice_hearing_details, without_notice_hearing_details, change_path: edit_steps_application_without_notice_details_path),
+        ].select(&:show?)
+      end
+
+      private
+
+      def without_notice_hearing_details
+        return [] unless c100.without_notice.eql?(GenericYesNo::YES.to_s)
```
zheileman: No need for this, the `AnswersGroup` will only render what is available, otherwise, will render nothing.

app/presenters/summary/html_sections/without_notice_details.rb
```diff
@@ -0,0 +1,30 @@
...
+
+        [
+          FreeTextAnswer.new(:without_notice_details,            c100.without_notice_details),
+          Answer.new(:without_notice_impossible,                 c100.without_notice_impossible),
+          FreeTextAnswer.new(:without_notice_impossible_details, c100.without_notice_impossible_details),
+          Answer.new(:without_notice_frustrate,                  c100.without_notice_frustrate),
+          FreeTextAnswer.new(:without_notice_frustrate_details,  c100.without_notice_frustrate_details),
+        ].select(&:show?)
```
zheileman: Also no need for the `select(&:show?)` here, because the `AnswersGroup` will do the filter for you inside the class.

PR #404, opened 2018-05-09
 Add children relationships section to people

zheileman:
Unfortunately it is just too much work to try to fit this into an AnswerGroup, as technically, it is not a single group. Each of the relations, require their own links, and that's just not possible with an AnswerGroup (the change links would not even appear in the right column, to start with).

So this is the second best approach.

screen shot 2018-05-09 at 11 51 32

### DIFF
--- a/app/presenters/summary/base_answer.rb
+++ b/app/presenters/summary/base_answer.rb
@@ -1,12 +1,17 @@
 module Summary
   class BaseAnswer
-    attr_reader :question, :value, :show, :change_path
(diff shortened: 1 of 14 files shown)


zheileman:
@aldavidson LOL yes, I named the children in a not very clever way: Child 1, Child 2, Other Child 1 and Other Child 2.
But yes, if you name the children properly, like John Doe, it will show "Relationship to John Doe" in the CYA.

PR #428, opened 2018-05-23
Attempt to make `wkhtmltopdf` play nice on Heroku

zheileman:
Something has changed in the Heroku stacks and we can't rely anymore on the `wkhtmltopdf` gem we were using.

As a first attempt, let's try a heroku-specific version.


zheileman:
This doesn't work.

PR #474, opened 2018-06-26
Add hull bristol and oxford courts

aldavidson:


zheileman:
Can we get rid of the 22 commits in this PR?

It should be a clean single commit if you branch from master which is up to date with all latest changes, I've done a test PR:

https://github.com/ministryofjustice/c100-application/pull/476

PR #493, opened 2018-07-18
prioritise emails marked as "Applications" over "Family"

aldavidson:
Primarily to accommodate the needs of Leeds Combined Court Centre under the new Courtfinder admin system


zheileman:
Hey Al, there is no such thing as `family applications` (I made this up in a previous PR, but it just doesn't exist). So it is completely safe to remove this condition. We didn't have that before anyway.

PR #538, opened 2018-11-05
Increase min pass length

zheileman:
Following ICO guidance on passwords for online services under GDPR.

### DIFF
--- a/spec/controllers/users/registrations_controller_spec.rb
+++ b/spec/controllers/users/registrations_controller_spec.rb
@@ -40,7 +40,7 @@ RSpec.describe Users::RegistrationsController do
       def do_post
         post :create, params: { 'user' => {
           'email' => 'foo@bar.com',
-          'password' => 'passw0rd'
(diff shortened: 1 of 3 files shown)


spec/controllers/users/registrations_controller_spec.rb
```diff
@@ -40,7 +40,7 @@
       def do_post
         post :create, params: { 'user' => {
           'email' => 'foo@bar.com',
-          'password' => 'passw0rd'
+          'password' => 'mypassw0rd'
```
solidgoldpig: What's the reason for the password change? Is the value irrelevant?

zheileman: It is a test, previously it passed with 8 chars as that was the minimum, now we've increased the minimum to 10 chars, so this test is also updated for that reason.

solidgoldpig: Doh! I couldn't see the value being checked for and overlooked the obvious requirement!

PR #584, opened 2019-01-21
Support for split children names

zheileman:
This first PR will set the basics to allow splitting full names into first / last name in the children names step. **More context and details in each of the individual commits**.

Once we are happy with the results and test there are no regressions we can extend the work to the rest of people (applicants, respondents, other parties ...)

For now, only C100 Application records with `version` greater than 1 will have the splitting of the children names enabled. This will allow us to release this in a controlled way as we need to ensure backward compatibility with existing saved applications in the database.

For testing purposes, in local, an ENV variable named `SPLIT_NAMES` can be set (to any value) to enable the splitting of the names regardless of the `version`.

**Note:** for some time we will have a `NamesBaseForm` and a `NamesSplitBaseForm` with some very similar code, and also some IF-ELSE here and there. Once all applications in the database are using the new split format, we can do a cleaning PR to remove no longer needed code and perhaps rename/refactor some other bits.

**Pending**: styling and perhaps copy changes.

screen shot 2019-01-21 at 13 37 33


app/models/person.rb
```diff
@@ -4,4 +4,13 @@ class Person < ApplicationRecord
 
   # Using UUIDs as the record IDs. We can't trust sequential ordering by ID
   default_scope { order(created_at: :asc) }
+
+  def full_name
+    [first_name, last_name].compact.join(' ')
+  end
+
+  # Temporarily until we've migrated all records to the new first/last name format
+  def full_name=(val)
```
teneightfive: What's this doing? Is it setting the full name to the last name value when called?

zheileman: Yes, as we repurposed the existing `name` (never really used) and `full_name` DB fields to be `first_name` and `last_name` respectively, and all the current forms were using `full_name`, instead of renaming everywhere this is a way of 'aliasing' any assignments of `full_name` to go the `last_name` field instead.

We will need this code for some time until all forms are using the split format first/last name, then we will not need it anymore and can be removed (part of a bigger cleaning up really).

PR #585, opened 2019-01-22
Extend split names to all parties

zheileman:
This will extend the work already done for children in previous PR #584 to all parties:

* applicant
* respondent
* other children
* other parties

Although it looks like A LOT of files, it actually is a very simple change, only multiplied by 4. Individual commits for easier read.

One could argue that's a lot of "duplicated" code, but in this case, given the life expectancy of most of this code will be less than 1 month (until all records in the database are using the new format) and then we are going to clean it up, it is not worth it coming up with another (more complicated) abstraction level.

### DIFF
--- a/app/views/steps/shared/_existing_names.html.erb
+++ b/app/views/steps/shared/_existing_names.html.erb
@@ -1,0 +1,8 @@
+<%= c.hidden_field :id %>
+
+<% if f.object.split_names? %>
+  <%= c.text_field :first_name, class: 'narrow' %>
+  <%= c.text_field :last_name, class: 'narrow' %>
+<% else %>
+  <%= c.text_field :full_name %>
+<% end %>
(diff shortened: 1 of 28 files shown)


app/views/steps/shared/_existing_names.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= c.hidden_field :id %>
```
teneightfive: What about using a longer format for `c` and `f` variables in these files?

When just looking at these files it won't be that obvious that they refer to `child` and `field` from the loop in the other view.

zheileman: Oh, yes, I just took what was already there (I have to admit in rails app is very common for this kind of form elements to be named just `f` or single letters) but out of the context makes it a bit obscure.

Will rename it and push.

teneightfive: Yeah I see if often and I think it's okay inside a loop like it was before when you can see the full reference nearby but here it feels like it's lost its descriptiveness.

PR #586, opened 2019-01-22
Update multiple names UI

teneightfive:
We now ask for both first name and last name in separate text fields
rather than one.

This could potentially lead to confusion when there are multiple names
as there isn't as good distinction between groups of names.

This change switches the font weight to create a visual separation
between the form groups.

## Before
localhost_3000_steps_children_names 2

## After
localhost_3000_steps_children_names 1

### DIFF
--- a/app/views/steps/other_children/names/edit.html.erb
+++ b/app/views/steps/other_children/names/edit.html.erb
@@ -9,8 +9,7 @@
     <%= step_form @form_object do |f| %>
       <% @existing_records.each.with_index do |child, index| %>
         <%= f.fields_for :names_attributes, child, index: index do |c| %>
-          <span><%= t('.record_index', index: index + 1) %></span>
-          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c } %>
+          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c, legend: t('.record_index', index: index + 1) } %>
         <% end %>
       <% end %>
 
--- a/app/views/steps/other_parties/names/edit.html.erb
+++ b/app/views/steps/other_parties/names/edit.html.erb
@@ -9,8 +9,7 @@
     <%= step_form @form_object do |f| %>
       <% @existing_records.each.with_index do |party, index| %>
         <%= f.fields_for :names_attributes, party, index: index do |c| %>
-          <span><%= t('.record_index', index: index + 1) %></span>
-          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c } %>
+          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c, legend: t('.record_index', index: index + 1) } %>
         <% end %>
       <% end %>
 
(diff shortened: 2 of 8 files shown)


app/views/steps/other_children/names/edit.html.erb
```diff
@@ -9,8 +9,7 @@
     <%= step_form @form_object do |f| %>
       <% @existing_records.each.with_index do |child, index| %>
         <%= f.fields_for :names_attributes, child, index: index do |c| %>
-          <span><%= t('.record_index', index: index + 1) %></span>
-          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c } %>
+          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c, t('.record_index', index: index + 1) } %>
```
zheileman: It needs the `legend` key:
```suggestion
          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c, legend: t('.record_index', index: index + 1) } %>
```

app/views/steps/other_parties/names/edit.html.erb
```diff
@@ -9,8 +9,7 @@
     <%= step_form @form_object do |f| %>
       <% @existing_records.each.with_index do |party, index| %>
         <%= f.fields_for :names_attributes, party, index: index do |c| %>
-          <span><%= t('.record_index', index: index + 1) %></span>
-          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c } %>
+          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c, t('.record_index', index: index + 1) } %>
```
zheileman: It needs the legend key:
```suggestion
          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c, legend: t('.record_index', index: index + 1) } %>
```

app/views/steps/respondent/names/edit.html.erb
```diff
@@ -13,8 +13,7 @@
     <%= step_form @form_object do |f| %>
       <% @existing_records.each.with_index do |respondent, index| %>
         <%= f.fields_for :names_attributes, respondent, index: index do |c| %>
-          <span><%= t('.record_index', index: index + 1) %></span>
-          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c } %>
+          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c, t('.record_index', index: index + 1) } %>
```
zheileman: It needs the legend key:
```suggestion
          <%= render partial: 'steps/shared/existing_names', locals: { names_form: f, person: c, legend: t('.record_index', index: index + 1) } %>
```

app/views/steps/shared/_existing_names.html.erb
```diff
@@ -1,8 +1,16 @@
 <%= person.hidden_field :id %>
 
-<% if names_form.object.split_names? %>
-  <%= person.text_field :first_name, class: 'narrow' %>
-  <%= person.text_field :last_name, class: 'narrow' %>
-<% else %>
-  <%= person.text_field :full_name %>
-<% end %>
+<fieldset class="fieldset-compact">
+  <% if legend %>
```
zheileman: I personally would get rid of this IF because in all instances we always pass a `legend` and I don't anticipate not having ever a legend in this partial, as it is only used in the names views.

teneightfive: I was thinking more for safety and then the possibility of using this partial for new names too.

I'll get rid of it for now though

PR #593, opened 2019-01-24
Fix focus state on button-secondary links

BNewing:
Some elements with the class of button-secondary are link tags, and when they are unvisited, their focus state made their text white and therefore unreadable against a whie background. This commit overrides that CSS by explicitely setting a button-secondary focus state text color.

### DIFF
--- a/app/assets/stylesheets/local/buttons.scss
+++ b/app/assets/stylesheets/local/buttons.scss
@@ -10,21 +10,40 @@
   cursor: pointer;
 }
 
-.button.button-secondary{
-  background:transparent;
-  color:#005ea5;
-  border-bottom-color:#fff;
-  box-shadow:none;
-  text-decoration:underline;
-  padding-left:0;
-  padding-right:0;
-  margin-left:96px;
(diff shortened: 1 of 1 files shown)


teneightfive:
Maybe include some before/after screenshots? I find it nicely helps when scanning the PR :)

app/assets/stylesheets/local/buttons.scss
```diff
@@ -25,6 +25,9 @@
   background:transparent;
   color:#2b8cc4;
 }
+.button.button-secondary:focus{
+  color:#005ea5;
```
teneightfive: I wonder if there's a variable we can use for this? Maybe try `$link-colour`.

teneightfive: We might be able to remove the double class definition too. So just define it as `.button-secondary:focus`. Or even just apply it to all elements with `.button` class?

BNewing: Added in the changes we made when pairing :)

PR #594, opened 2019-01-24
Realign checkboxes in MIAM section

BNewing:
The gem updates made the checkboxes slightly out of alignment on the custom MIAM page. This commit fixes the CSS so that the text now lines up with the checkbox. It also removes <br> elements from the text, which were no longer needed and also causing layout issues because the <span> elements had been given display: block in the gem update.

## Before:
image

## After:
screen shot 2019-01-24 at 15 24 41

### DIFF
--- a/app/assets/stylesheets/local/forms.scss
+++ b/app/assets/stylesheets/local/forms.scss
@@ -1,26 +1,32 @@
-form.button_to {
-  padding-bottom: 0;
+form {
+  .app_heading {
+      margin-bottom: 3rem;
+  }
+  
+  &.button_to {
+    padding-bottom: 0;
 
-  #proposition-links & {
-    input[type="submit"] {
-      background: transparent;
-      border: none;
-      outline: none;
-      color: $white;
-      padding: 0;
-      margin: 0;
-      font-weight: bold;
-      cursor: pointer;
-      text-decoration-skip: ink;
+    #proposition-links & {
+      input[type="submit"] {
+        background: transparent;
+        border: none;
+        outline: none;
+        color: $white;
+        padding: 0;
+        margin: 0;
+        font-weight: bold;
+        cursor: pointer;
+        text-decoration-skip: ink;
 
-      &:hover {
-        text-decoration: underline;
-      }
+        &:hover {
+          text-decoration: underline;
+        }
 
-      &:focus {
-        color: $black;
-        background-color: $yellow;
-        outline: 3px solid $yellow;
+        &:focus {
+          color: $black;
+          background-color: $yellow;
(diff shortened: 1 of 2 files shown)


config/locales/exemptions/en.yml
```diff
@@ -130,14 +130,14 @@ en:
 
     MISC_EXEMPTIONS: &MISC_EXEMPTIONS
       no_respondent_address_html: |
-        <span class="gv-u-heading-small">You don’t have sufficient contact details for the other people in this application (the respondents)</span><br/>
+        <span class="gv-u-heading-small">You don’t have sufficient contact details for the other people in this application (the respondents)</span>
         <span class="form-hint">The other people can’t be contacted to arrange a MIAM.</span>
       without_notice_html: |
         <span class="gv-u-heading-small">You’re applying for a without notice hearing</span><br/>
```
zheileman: I think we can get rid of this `<br/>` too and that would be all of them 😃 
```suggestion
         <span class="gv-u-heading-small">You’re applying for a without notice hearing</span>
```

BNewing:
## Before:
image

## After:
screen shot 2019-01-24 at 15 24 41

app/assets/stylesheets/local/forms.scss
```diff
@@ -1,62 +1,62 @@
...
-      &:hover {
-        text-decoration: underline;
-      }
-
-      &:focus {
-        color: $black;
-        background-color: $yellow;
-        outline: 3px solid $yellow;
+form {
+  & .app_heading {
```
teneightfive: We don't need the `&` here. As `.app_heading` is a child we can define it without needing to use the parent selector using the `&` symbol

app/assets/stylesheets/local/forms.scss
```diff
@@ -83,35 +83,37 @@ label[for="steps_abduction_risk_details_form_risk_details"]
...
   }
 }
 
 .form-label {
   font-weight: bold;
-}
 
-.form-label-normal {
-  font-weight: normal;
+  &-normal {
```
teneightfive: I would probably avoid joining this with an `&` and keep them separate. It can make it more confusing to find that particular selector using a search.

app/assets/stylesheets/local/forms.scss
BNewing: Changes addressed in a fixup commit :)

app/assets/stylesheets/local/forms.scss
BNewing: Changes also addressed in a fixup commit :)

PR #599, opened 2019-02-01
Make `remove name` work without JS

zheileman:
We had a functionality depending on JS being enabled. This is not ideal and we've refactor the code to get rid of the JS dependency and as a side effect enable the possibility of remove individual names, no matter their position in the list.

As a consequence of this change, there is now some styling TLC and possibly the need to disable the ENTER key with JS, which is a minor inconvenience.

screen shot 2019-02-01 at 11 21 20

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -448,4 +448,4 @@ RUBY VERSION
    ruby 2.5.1p57
 
 BUNDLED WITH
-   1.17.0
+   1.17.3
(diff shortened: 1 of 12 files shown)


teneightfive:
Just pushed 2 commits which change it to look like this:

screenshot 2019-02-01 at 13 47 38

@zheileman / @BNewing what do you think?

PR #601, opened 2019-02-04
Disable `enter` keypress in names form

zheileman:
A side effect of having individual 'remove' buttons is, on `enter` keypress in any of the text inputs, this submit action is actioned, removing the active name.

In order to stop this behaviour, we are going to disable form submits on enter keypress, only on text inputs, and only for the names form, not other forms.

### DIFF
--- a/app/assets/javascripts/modules/disable-enter-key.js
+++ b/app/assets/javascripts/modules/disable-enter-key.js
@@ -1,0 +1,18 @@
+'use strict';
+
+moj.Modules.disableEnterKey = {
+  init: function() {
+    var self = this;
+    self.bindEvents();
+  },
+
+  bindEvents: function() {
+    $("form[data-disable-enter] input[type='text']").on('keyup keypress', function(e) {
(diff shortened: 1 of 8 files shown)


solidgoldpig:
@zheileman @BNewing May I suggest a better solution? Add a default submit input(that does the same as your primary action) to the beginning of the form. Set that input to display: none.

zheileman:
@solidgoldpig I see what you mean, problem is, there is no clear expectation of what we should do here. In other forms, enter key will submit the form, and if there are no validation errors, it means it will "advance" to the next step.

But in the names steps, it is a bit more complicated, as if we just say "let's advance" that means a user filling a first/last name, and hitting `enter` in the last name, will advance the form to the next step, which might not be ideal as they might want to add more children, or more people in general.

If instead of advance we say, ok, "let's just add this name and reload the page" then we are back to the original "bug report" (_When entering your child's names, if you hit return in the "Childs Name" box, it doesn't Continue - it adds another child_)

PR #616, opened 2019-02-27
Add legend text to radio buttons

BNewing:
Some of the radio buttons don't currently have the question in the legend. This addresses that so that for screen reader users, the question is directly associated with the input field.

## What

[Link to story](https://mojdigital.teamwork.com/index.cfm#/tasks/14030794)

@zheileman has updated the gem, so that text can be added to radio button legends. This PR goes through the individual views so that they use this new functionality, and removes now unnecessary strings from the YAML.

## Checklist

### DIFF
--- a/app/views/steps/miam/certification_date/edit.html.erb
+++ b/app/views/steps/miam/certification_date/edit.html.erb
@@ -20,5 +20,3 @@
     <% end %>
   </div>
 </div>
-
-# TODO: Sort out making the hint text visually unhidden when the legend is visually hidden
--- a/app/views/steps/miam/certification_date/edit.html.erb
+++ b/app/views/steps/miam/certification_date/edit.html.erb
@@ -20,3 +20,5 @@
     <% end %>
   </div>
 </div>
+
+# TODO: Sort out making the hint text visually unhidden when the legend is visually hidden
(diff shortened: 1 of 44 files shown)


app/views/steps/shared/_relationship_step.html.erb
```diff
@@ -6,7 +6,7 @@
 
 <%= step_form form_object do |f| %>
   <%=
-    f.radio_button_fieldset :relation do |fieldset|
+    f.radio_button_fieldset :relation, { legend_options: { class: 'visually-hidden', text: t('.heading', person_name: form_object.record.person.full_name, child_name:  form_object.record.minor.full_name) } } do |fieldset|
```
zheileman: I think in cases where we need to pass arguments to `t()` it might be better and cleaner to declare a variable once, and reuse it in the header and in the legend.

For example, in the top of the page:

```ruby
heading = t('.heading', person_name: form_object.record.person.full_name, child_name: form_object.record.minor.full_name)
```

And then we use `heading` variable in the `<h1>` and in the `legend_options `

app/views/steps/miam/certification_date/edit.html.erb
```diff
@@ -20,3 +20,5 @@
     <% end %>
   </div>
 </div>
+
+# TODO: Sort out making the hint text visually unhidden when the legend is visually hidden
```
zheileman: We can't leave this comment here, as it will be visible in the page.
We can change it to an HTML comment `<!-- blah --> `

app/views/steps/shared/_relationship_step.html.erb
BNewing: Good call! I have added this here and into other places where the heading was more than just '.heading'

app/views/steps/miam/certification_date/edit.html.erb
BNewing: I have removed this :)

PR #621, opened 2019-03-05
Tidy up checkbox legends

BNewing:
## What

[Link to story](https://mojdigital.teamwork.com/index.cfm#/tasks/14567208)

Some of the checkboxes don't need legends, so this removes some of the unneccessary yml strings for them. It also adds in legends for checkbox fieldsets that need them.

## Checklist

- [ ] Tests and rubocop should be passing: `bundle exec rake`
- [ ] Github should not be reporting conflicts; you should have recently run `git rebase master`.
- [ ] There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- [ ] The PR description should say what you changed and why, with a link to the story.
- [ ] You should have looked at the diff against master and ensured that nothing unexpected is included in your changes.
- [ ] You should have checked that the commit messages say why the change was made.

### DIFF
--- a/app/views/steps/petition/orders/edit.html.erb
+++ b/app/views/steps/petition/orders/edit.html.erb
@@ -6,33 +6,17 @@
 
     <h1 class="heading-xlarge gv-u-heading-xxlarge"><%=t '.heading' %></h1>
 
-    <p class="lede gv-u-text-lede"><%=t '.lead_text' %></p>
-
     <%= step_form @form_object do |f| %>
-      <%= f.check_box_fieldset :orders, PetitionOrder::CHILD_ARRANGEMENTS %>
-
       <%=
-        f.check_box_fieldset :orders, [
-          PetitionOrder::PROHIBITED_STEPS,
-        ] do |fieldset|
+        f.check_box_fieldset :orders, [], legend_options: { class: 'visually-hidden', text: t('.heading') } do |fieldset|
+          fieldset.check_box_input(PetitionOrder::CHILD_ARRANGEMENTS_HOME)
+          fieldset.check_box_input(PetitionOrder::CHILD_ARRANGEMENTS_TIME)
           fieldset.check_box_input(PetitionOrder::GROUP_PROHIBITED_STEPS) {
             f.check_box_fieldset :orders_prohibited_steps, PetitionOrder::PROHIBITED_STEPS
           }
-        end
-      %>
-
-      <%=
-        f.check_box_fieldset :orders, [
-          PetitionOrder::SPECIFIC_ISSUES,
-        ] do |fieldset|
           fieldset.check_box_input(PetitionOrder::GROUP_SPECIFIC_ISSUES) {
             f.check_box_fieldset :orders_specific_issues, PetitionOrder::SPECIFIC_ISSUES
           }
-        end
-      %>
-
-      <%=
-        f.check_box_fieldset :orders, [PetitionOrder::OTHER_ISSUE] do |fieldset|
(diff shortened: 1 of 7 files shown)


BNewing:
So far, I've gone through and done the easy ones. There are a few issues left, which need to be addressed, and as this PR is quite small at the moment I'd say we're okay to add them onto this one. The outstanding issues are:

- [ ] anything in the yml file in the fieldset section that has 'exemptions_group_html' in it is where several fieldsets are nested, which is not ideal
- [ ] the petition/orders page has several fieldsets next to each other instead of one big fieldset like there ought to be
- [ ] there are nested fieldsets on abuse_concerns/details
- [ ] on respondent/names and children/names, 'don't know' should either be clearer about what it relates to or nested in with the fieldset for the date question, which it relates to - or if possible both

BNewing:
- 'make sure the design is consistent, and that 'select that all that apply' looks the same across the whole form'

To check this, I did a text editor search for the checkbox, and then took a screenshot of all instances it in the application. I then went through the screenshots and compared them. In doing so, I found the following. miam_exemption pages follow this pattern:

image

This follows the GDS pattern so is okay. There is another pattern, which is without the fieldset, which is follows GDS guidance, and is also okay: 

image

However petition/orders is slightly different:

image

The issue here is that the 'select all that apply' text is not a form hint, but a normal `<p>` tag that lives outside of the form. This is a knock on issue from that form being made up of several `<fieldset>` elements whilst all being about the same thing. 

This will be addressed as part of that issue, so I'm going to mark "make sure the design is consistent, and that 'select that all that apply' looks the same across the whole form" as done.

app/views/steps/petition/orders/edit.html.erb
```diff
@@ -6,33 +6,17 @@
...
-
     <%= step_form @form_object do |f| %>
-      <%= f.check_box_fieldset :orders, PetitionOrder::CHILD_ARRANGEMENTS %>
-
       <%=
-        f.check_box_fieldset :orders, [
-          PetitionOrder::PROHIBITED_STEPS,
-        ] do |fieldset|
+        f.check_box_fieldset :orders, [], legend_options: { class: 'visually-hidden', text: t('.heading') } do |fieldset|
+          fieldset.check_box_input(PetitionOrder::CHILD_ARRANGEMENTS_HOME)
```
BNewing: Is there a way of flattening this all out so that we're not rendering more `<fieldset>` elements? I tried replacing the `fieldset.check_box_input` with `f.check_box_input` but that wasn't quite enough

zheileman: No. It is the same in the MIAM exemptions.

There is a top level fieldset, and inside there are as many other fieldsets as groups of checkbox we have (those ones that open when you check a checkbox).

So here we have 2 groups, there will be 2 fieldsets (one for `orders_prohibited_steps` and another one for `orders_specific_issues`), inside the top level fieldset.

It is what creates the "group" of checkboxes. And we have 3. One top level with the first-level checkboxes, and then 2 sub-groups.

BNewing:
After a bit more research, it's unclear whether the last two points on the list are issues. Leonie Watson, a screen reader user, says that nested fieldsets cause issues for screen reader users (http://test-cases.tink.uk/nested-fieldsets/). However the test case in this example doesn't map directly to ours. We would need to test with screen reader users to really know if there is an issue and how to address it here in this particular pattern. 

As a result, we're avoiding making any further changes, because without a clear plan of improvements , there's a risk of making the experience worse.

PR #623, opened 2019-03-06
Update page titles

BNewing:
## What

[Link to story](https://mojdigital.teamwork.com/index.cfm#/tasks/14567479)

This will update the page titles to the new content supplied by the Content Designer. Previously, the page titles weren't always clear about the purpose of the page, and the page title is the first thing read out to screen reader users when a new page loads. So this clarifies the pages to them.

The content has been taken from this Google spreadsheet from Amanda (Content Designer):
https://docs.google.com/spreadsheets/d/1RbZ3Lke3zfY0wkxNBdbw73647F7PEmMChB54XU65ffo/edit#gid=0

## Checklist

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -205,7 +205,7 @@ en:
           body_info: Include as much detail as you can. This will help us process your application more quickly.
       has_other_parties:
         edit:
-          page_title: Other parties required
+          page_title: Other parties who should know about application
           heading: Is there anyone else who should know about your application?
           info_html: |
             <p>For example, you should tell everyone who:</p>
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -205,7 +205,7 @@ en:
           body_info: Include as much detail as you can. This will help us process your application more quickly.
       has_other_parties:
         edit:
-          page_title: Other parties required
+          page_title: Other parties who should know about application
```
zheileman: Is a `the` missing here? "about the application"

BNewing: I've just double checked and it's the recommendation Amanda's made and it's a direct copy and paste - is it worth us confirming with her when she's back?

PR #648, opened 2019-05-07
Heroku to MOJ Cloud Platforms data migration

zheileman:
Minimal docker image (base image `node:alpine`) where we install Heroku CLI and postgres client in order to run a series of scripts.

The plan is to create a one-off Job or Pod in the corresponding namespace (staging or production) and use this docker image with access to the necessary env variables.

* backup.sh -- will place the app in maintenance mode and take a database dump of a Heroku application (`HEROKU_APP`)

* restore.sh -- will restore the previous db dump to another database, with connection string `DATABASE_URL`.

If `DRY_RUN` env variable is defined, will skip destructive commands.

A `HEROKU_API_KEY` env variable is needed, with access to the heroku app.

### DIFF
--- a/heroku-migration/Dockerfile
+++ b/heroku-migration/Dockerfile
@@ -1,0 +1,19 @@
+FROM node:alpine
+
+ENV WORKDIR /usr/src/app
+WORKDIR $WORKDIR
+
+COPY scripts scripts/
+
+RUN apk add --no-cache bash curl postgresql-client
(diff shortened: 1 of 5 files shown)


zheileman:
We've now successfully migrated to cloud platforms.
Closing this PR (as there is no need to merge this, we are not going to run it again).

I've also closed [associated PR](https://github.com/ministryofjustice/c100-application-deploy/pull/14) in the deploy repo.

PR #646, opened 2019-05-02
Add split text boxes for address for applicant address details form

tmckernan:
## What

[Link to story](https://mojdigital.teamwork.com/index.cfm#/tasks/15152297)

- Add address_data JSON column to the Person model.
- Add feature flag SPLIT_ADDRESS
- Update applicant address form to separate address column into several text boxes and save in address_data form, when feature flag SPLIT_ADDRESS is enabled.

New applicant  address details form look like the screenshot below:

Screen Shot 2019-05-02 at 12 23 55

## Checklist

Before you ask people to review this PR:

- [x] Tests and rubocop should be passing: `bundle exec rake`
- [x] Github should not be reporting conflicts; you should have recently run `git rebase master`.
- [x] There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- [x] The PR description should say what you changed and why, with a link to the story.
- [x] You should have looked at the diff against master and ensured that nothing unexpected is included in your changes.
- [x] You should have checked that the commit messages say why the change was made.

PR #653, opened 2019-05-10
Add split text boxes for other parties address details form

tmckernan:
## What

[Link to story](https://mojdigital.teamwork.com/index.cfm#/tasks/15152297)

Update other parties address form to separate address column into several text boxes and save in address_data form, when feature flag SPLIT_ADDRESS is enabled.  This is a follow on from [#646](https://github.com/ministryofjustice/c100-application/pull/646) [#647] and (https://github.com/ministryofjustice/c100-application/pull/647) PRs only for other parties.

## Checklist

### DIFF
--- a/app/models/other_party.rb
+++ b/app/models/other_party.rb
@@ -1,5 +1,4 @@
 class OtherParty < Person
-  def full_address
-    address
-  end
+  store_accessor :address_data, :address_line_1, :address_line_2,
+                 :town, :country, :postcode
 end
--- a/app/models/other_party.rb
+++ b/app/models/other_party.rb
@@ -1,4 +1,2 @@
 class OtherParty < Person
-  store_accessor :address_data, :address_line_1, :address_line_2,
-                 :town, :country, :postcode
 end
--- a/app/views/steps/other_parties/address_details/edit.html.erb
+++ b/app/views/steps/other_parties/address_details/edit.html.erb
@@ -7,7 +7,15 @@
     <h1 class="heading-xlarge gv-u-heading-xxlarge"><%=t '.heading', name: @form_object.record.full_name %></h1>
 
     <%= step_form @form_object do |f| %>
-      <%= f.text_area :address, size: '40x4', class: 'form-control-3-4' %>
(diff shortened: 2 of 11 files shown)


spec/forms/steps/other_parties/address_details_form_spec.rb
```diff
@@ -71,5 +89,89 @@
...
+          address_line_1: address_line_1,
+          address_line_2: address_line_2,
+          town: town,
+          country: country,
+          postcode: postcode
+        }
+      end
+
+      context 'validations on field presence unless `unknown`' do
+        it { should validate_presence_of(:address_line_1) }
```
zheileman: Shouldn't this be:

```suggestion
        it { should validate_presence_unless_unknown_of(:address_line_1) }
```

Same for `postcode`.

Also, revisit the respondent spec in case the same has happened.

app/views/steps/other_parties/address_details/edit.html.erb
```diff
@@ -7,7 +7,15 @@
     <h1 class="heading-xlarge gv-u-heading-xxlarge"><%=t '.heading', name: @form_object.record.full_name %></h1>
 
     <%= step_form @form_object do |f| %>
-      <%= f.text_area :address, size: '40x4', class: 'form-control-3-4' %>
+      <% if f.object.split_address? %>
+          <%= f.text_field :address_line_1 %>
```
zheileman: Sorry to be that guy LOL 🤣 indentation is off (I think it is also off in the other address details views and this is a copy&paste).

app/models/other_party.rb
```diff
@@ -1,5 +1,4 @@
 class OtherParty < Person
-  def full_address
-    address
-  end
+  store_accessor :address_data, :address_line_1, :address_line_2,
```
zheileman: Let's move now this accessor (in applicant, respondent and other_party) to the super class `Person` so everybody inherits it (I know we don't use addresses for children/minors but it is ok to have the accessor, as they also inherits the rest of the table attributes regardless of being used or not in form objects).

PR #657, opened 2019-05-14
Add post code address lookup service

tmckernan:
## What

[Link to story](https://mojdigital.teamwork.com/index.cfm#/tasks/15321247)

A service with accepts a uk post code and returns a list of addresses for that given post code.  The service uses the ordnance survey API to perform the post code look up.  Return an empty array if no addresses or found for a given postcode.

## Checklist


app/services/c100_app/address_lookup_errors.rb
```diff
@@ -0,0 +1,28 @@
+module C100App
+  class NotImplementedError < ::StandardError; end
```
zheileman: Unused, can be removed.

spec/rails_helper.rb
```diff
@@ -13,6 +13,8 @@
 RSpec.configure do |config|
   config.infer_spec_type_from_file_location!
 
+  config.fixture_path = "#{::Rails.root}/spec/fixtures"
```
zheileman: Not needed, as long as we use the default path.

app/services/c100_app/address_lookup_service.rb
```diff
@@ -0,0 +1,64 @@
+module C100App
+  class AddressLookupService
+    ORDNANCE_SURVEY_URL = 'https://api.ordnancesurvey.co.uk/places/v1/addresses/postcode'.freeze
+
+    attr_accessor :postcode, :result
+
+    def initialize(postcode)
+      @postcode = postcode
+      @result = MapAddressLookupResults.call(results)
```
zheileman: I think if you remove this from the `#initialize`, it will be easier to test. So one step is initialising the service, and another step is calling `#results` to get the results (which, if we don't have them, will trigger the API call).

Otherwise there is a complex chain of events (init -> call API -> map results from API)

app/services/c100_app/address_lookup_service.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    def errors
+      @errors ||= AddressLookupErrors.new
+    end
+
+    private
+
+    def query_params
+      {
+        key: ENV.fetch('ORDNANACE_SURVEY_API_KEY'),
```
zheileman: In the tests, we must always mock a default test value for this, otherwise it is not possible to run the tests locally (and I imagine same on CI) unless the env variable is declared:

```ruby
Failure/Error: key: ENV.fetch('ORDNANACE_SURVEY_API_KEY'),
     
     KeyError:
       key not found: "ORDNANACE_SURVEY_API_KEY"
     # ./app/services/c100_app/address_lookup_service.rb:24:in `fetch'
```

PR #667, opened 2019-05-21
Ensure a predictable full address order

zheileman:
The order in the `address_data` Hash (store) is not guaranteed, and in fact, some keys (postcode or country) are created before others (address lines).

This has the side effect of sometimes returning a funny `full_address` that does not maintain the order people are used to.

Make the `#full_address` method resilient to these scenarios and always return the attribute values in the same order they were declared, regardless of the order of the keys in the Hash.

## Checklist

### DIFF
--- a/app/models/person.rb
+++ b/app/models/person.rb
@@ -1,8 +1,9 @@
 class Person < ApplicationRecord
-  store_accessor :address_data, :address_line_1, :address_line_2,
-                 :town, :country, :postcode
   belongs_to :c100_application
   has_many :relationships, source: :person, dependent: :destroy
+
+  store_accessor :address_data,
+                 :address_line_1, :address_line_2, :town, :country, :postcode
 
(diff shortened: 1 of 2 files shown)


zheileman:
> LGTM if there is any cases when the address_data Hash (store) isn't in the right order
> 
> I don't think there will be any cases in which the address_data Hash (store) isn't in the correct order as we set the order in
> 
> https://github.com/ministryofjustice/c100-application/blob/418a70ac4a10b753b0a814a1bacae87e5f05a0e5/app/forms/address_base_form.rb#L45

Yes but before reaching that point we've had to store the postcode from the lookup (and also we are defaulting the country to UK) so those 2 keys are stored before the rest, and provoke this side effect.

It is very easy to alter the order in the `#persist` / `#update` methods (they are not intended to force a specific order in the attributes) so I think it is better not to rely on these, and be explicit about the order in the method that returns the value (in this case builds the full address string).

PR #818, opened 2020-03-06
Add an application fulfilment custom validator

zheileman:
This validator will run only on completion of the application (still not wired up, that will come up in a follow-up PR) and will check some preconditions are met in order for the application to be considered "valid" and to be submitted to court. Essentially a quick sanity check.

This is not intended to be an in-deep validation of every single attribute (as that already happens at the step level, one by one as the applicant progress through the application) but more for cases where the user jumped some steps or deleted by mistake all applicants, or all respondents, etc.

Still to do in a follow-up PR the wiring to the CYA page, and the presentation of the errors using the `FulfilmentErrorsPresenter` class and the `check_your_answers/fulfilment_error` partial (not included in this PR as we are awaiting design).

Note: as there is nothing yet on the service to "see" this working, apart from the tests, this can also be tested quickly by going to the rails console and running:

```ruby
a = C100Application.new
a.valid?(:completion)
FulfilmentErrorsPresenter.new(a).errors
```

### DIFF
--- a/spec/presenters/fulfilment_error_spec.rb
+++ b/spec/presenters/fulfilment_error_spec.rb
@@ -1,0 +1,31 @@
+require 'spec_helper'
+
+RSpec.describe FulfilmentError do
+  subject {
+    described_class.new(:foobar, message: 'Foobar message', error: :blank, change_path: 'steps/foo/bar')
+  }
+
+  describe '#to_partial_path' do
+    it 'returns the correct partial path' do
+      expect(subject.to_partial_path).to eq('steps/application/check_your_answers/fulfilment_error')
+    end
+  end
+
+  context 'expose the error details as attributes' do
+    describe '#attribute' do
+      it { is_expected.to have_attributes(attribute: :foobar) }
+    end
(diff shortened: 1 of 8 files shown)


spec/presenters/fulfilment_error_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    described_class.new(:foobar, message: 'Foobar message', error: :blank, change_path: 'steps/foo/bar')
+  }
+
+  describe '#to_partial_path' do
+    it 'returns the correct partial path' do
+      expect(subject.to_partial_path).to eq('steps/application/check_your_answers/fulfilment_error')
+    end
+  end
+
+  describe 'attributes accessors' do
```
Davidslv: I think you maybe have mistakenly used `describe` where you should be using `context` and vice versa

zheileman: Yep, in this scenario changing describe with context and vice versa will probably read better. Will push an update!

PR #819, opened 2020-03-09
Present fulfilment errors in CYA page

zheileman:
Follow-up to PR #818.

Show the errors, if any, in the CYA page. Still design is TBD so the partials are just placeholders, showing basic details about the error.

As part of this PR and to protect this new feature from being released to production before is ready, added a simple mechanism to feature-flag the validations: it will run ONLY if we are using fake court data (a fixture, like we do when ue the bypass functionality on local/staging envs).

This way, on production there is no risk of this validation being run just yet.

Note: it is possible to avoid the validations also on local/staging if you start the screener and answer at least one question, then you can use the bypass functionality (or just continue all the way to the end answering questions) and the fulfilment validator will not trigger.

Screen Shot 2020-03-09 at 13 23 45

### DIFF
--- a/spec/forms/steps/application/declaration_form_spec.rb
+++ b/spec/forms/steps/application/declaration_form_spec.rb
@@ -23,18 +23,53 @@ RSpec.describe Steps::Application::DeclarationForm do
     end
 
     context 'validations' do
+      before do
+        allow(c100_application).to receive(:valid?)
+      end
+
       it { should validate_presence_of(:declaration_signee) }
       it { should validate_presence_of(:declaration_signee_capacity, :inclusion) }
     end
 
-    context 'when form is valid' do
-      it 'saves the record' do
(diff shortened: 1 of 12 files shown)


spec/forms/steps/application/declaration_form_spec.rb
```diff
@@ -23,18 +23,53 @@
...
-        ).and_return(true)
+    context 'when application fulfilment is successful' do
+      before do
+        allow(c100_application).to receive(:valid?).with(:completion).and_return(true)
+      end
+
+      context 'and form is valid' do
+        it 'saves the record' do
+          expect(c100_application).to receive(:update).with(
+            declaration_signee: 'Full Name',
```
Davidslv: it seems to me that you can use the values of the `let`'s on line 11 and 12 to fill the values here.

PR #822, opened 2020-03-11
add under_age boolean column on people database table

Davidslv:
I'm adding the column so that it can be stored over the user journey as we
would like to allow users under 18 to fill the form, trough a slightly different
journey that will show them information that they need a "litigation friend".

The implementation will come in a separate PR.

Story: https://mojdigital.teamwork.com/#/tasks/19603602


Davidslv:
merging

PR #827, opened 2020-03-13
Update under age page copy

Davidslv:
This pull request depends on https://github.com/ministryofjustice/c100-application/pull/822 to be merged first.

This commit replaces the copy and adds information about which court the person
needs to get in contact with.

image

Screenshot 2020-03-13 at 15 00 30


Davidslv:
@zheileman I've tested the assignment of `@court` instance variable, I've done it in a separate commit so you can see it in isolation here: https://github.com/ministryofjustice/c100-application/pull/827/commits/5cae60b8469a0619b0de082694af515fac03a4c7

PR #828, opened 2020-03-16
Customise email copy when applicant is under age

zheileman:
Expose a helper method on the `Applicants` collection to query if there is at least one applicant under age (returning true or false).

This method is later used on Notify emails, but can also be used for the confirmation pages or any other place where we need to know if the application as a whole contains at least one under age.

Send the yes/no to Notify to be able to show/hide the instructions when applicant is under age in the confirmation email.

Tweaked a bit the card payment instructions as per suggestions from content designer.
The under18s instructions are part of the Notify template.

Example of email with the under 18s code inserted (and the payment instructions tweaked):

Screen Shot 2020-03-16 at 12 41 07

### DIFF
--- a/app/models/applicant.rb
+++ b/app/models/applicant.rb
@@ -1,2 +1,7 @@
 class Applicant < Person
+  # In the case of more than one applicant, if at least one of them
+  # is under age, the whole application is considered as "under age"
+  def self.under_age?
+    any?(&:under_age)
+  end
 end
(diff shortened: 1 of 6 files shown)


app/models/applicant.rb
```diff
@@ -1,2 +1,7 @@
 class Applicant < Person
+  # In the case of more than one applicant, if at least one of them
+  # is under age, the whole application is considered as "under age"
+  def self.under_age?
```
Davidslv: 👍 I like that, thanks!

PR #832, opened 2020-03-17
remove over18 attribute from screen answers

Davidslv:
This is because there is validation on the table attributes,
if those attributes are not filled, the screener never reaches the "completion" status that is needed.

### DIFF
--- a/db/migrate/20200317155024_remove_over18_from_screener_answers.rb
+++ b/db/migrate/20200317155024_remove_over18_from_screener_answers.rb
@@ -1,0 +1,5 @@
+class RemoveOver18FromScreenerAnswers < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :screener_answers, :over18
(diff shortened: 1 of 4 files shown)


db/migrate/20200317155024_remove_over18_from_screener_answers.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveOver18FromScreenerAnswers < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :screener_answers, :over18
```
zheileman:
```suggestion
    remove_column :screener_answers, :over18, :string
```

Just a very minor thing. Although we don't plan to get back and recover this attribute, the `change` method will need to now the attribute type for the case of a `down` (rollback). So adding the `string` type is recommended.

Davidslv:
👍

PR #837, opened 2020-03-24
Bring over GovukComponents code from Disclosure Checker

zheileman:
For the Disclosure Checker service we already managed to migrate some components although in a quite hacky way. For reference:

ministryofjustice/disclosure-checker#4

Although it is not the best of the code and it is a bit convoluted, it will be a great way to kick off the migration of components for C100 and speed up the process.

Also, as opposite to Disclosure Checker were we have "simple" components, here we make use of more complex components like revealing radios, etc. so that will need to be adapted too.

The idea is once we have a more clear understanding of the complexity of this, we can organise the components into individual classes and reuse as much as possible as well as eventually extract all the code to a gem that could also be used by Disclosure Checker.

For now, only the error summary and the text field have been migrated.

Migrated the first step (children's postcode) (refer to individual commit):

Screen Shot 2020-03-24 at 09 53 12

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -18,24 +18,28 @@ module ApplicationHelper
 
   # Render a back link pointing to the user's previous step
   def step_header(path: nil)
+    # TODO: remove the error summary from here once we've migrated all the views
+    content_for(:old_error_summary, &method(:error_summary))
+
     capture do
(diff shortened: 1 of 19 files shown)


zheileman:
Closing this PR as not viable.

We will continue working in the migration but making use of DfE gem (and submitting PRs and improvements where the gem lacks some functionality we need).

PR #872, opened 2020-04-16
Migrate fourth page of MIAM exemptions (ADR)

zheileman:
After running this through our service designer, we've simplified the page so instead of 2 check boxes revealing 3 and 2 each more, now it is just a plain list of check boxes, identical to the paper form.

This let's us remove some of the ugly hint text with bullet points, and make the page simpler to read.

Screen Shot 2020-04-16 at 14 45 24

### DIFF
--- a/app/views/steps/miam_exemptions/adr/edit.html.erb
+++ b/app/views/steps/miam_exemptions/adr/edit.html.erb
@@ -1,34 +1,29 @@
 <% title t('.page_title') %>
+<% step_header %>
 
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <%= step_header %>
-    <p class="app__section_heading"><%=t '.section' %></p>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary %>
 
-    <h1 class="app__heading heading-xlarge gv-u-heading-xxlarge"><%=t '.heading' %></h1>
+    <span class="govuk-caption-xl"><%=t '.section' %></span>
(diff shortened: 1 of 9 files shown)


app/views/steps/miam_exemptions/adr/edit.html.erb
```diff
@@ -1,34 +1,29 @@
...
+    <%= govuk_error_summary %>
+
+    <span class="govuk-caption-xl"><%=t '.section' %></span>
+    <h1 class="govuk-heading-xl"><%=t '.heading' %></h1>
+
+    <%= step_form @form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_check_boxes_fieldset :adr, legend: { tag: 'span', size: 'm' } do %>
+
+        <%= f.govuk_check_box :adr, AdrExemptions::PREVIOUS_ADR[0].to_s, link_errors: true %>
+        <%= f.govuk_check_box :adr, AdrExemptions::PREVIOUS_ADR[1].to_s %>
```
Davidslv, line 15: Not a blocker, if you don't care about their positions after the first one, you could do something similar to

```ruby
<% check_values = AdrExemptions::PREVIOUS_ADR[1..2] + AdrExemptions::PREVIOUS_MIAM %>

<% check_values.each do |value| %>
   <%= f.govuk_check_box :adr, value.to_s %>
<% end %>
```

zheileman: Yep! it makes sense in this case. Will change it.

PR #875, opened 2020-04-20
Migrate remainining exemptions pages (playback, kickout)

zheileman:
Adapt the playback pages to take the correct locales (without the `_html` suffix now) after the refactor of the MIAM exemptions checkbox pages.

Move/tweak the styles that generates numbered icon lists to the custom `app.scss` stylesheet. There are a few more pages using the old styles, will be updated in another commit.


app/views/steps/completion/what_next/show.en.html.erb
```diff
@@ -13,7 +13,9 @@
       </strong>
     </div>
 
-    <section class="numbered-list">
+    <!-- TODO: remove the `section` wrapper when updating this view to new design system
```
Davidslv, line 16: is this left intentionally here?

zheileman: Yes, I added that comment to not forget when we migrate the completion page, to remove the wrapper around the list (section) and to move the `app-list--icon-steps` class into the list.
I didn't want to do the migration of this page as part of this PR.

PR #876, opened 2020-04-20
Migrate nature of application (petition order) view

zheileman:
Migrate this step and its playback to new design system.

This is very similar to what was done for MIAM Exemptions, but it has a few particular differences so it is not worth trying to adapt the concern just for this (it has a revealing text area, for once), so I've opted for leaving the logic as it was, inside the form object itself, but updated to work with the new check boxes.

Screen Shot 2020-04-20 at 16 15 43


zheileman:
The protection orders page is a simple one so I've added it to this PR too, in commit 4418906

Screen Shot 2020-04-21 at 09 54 38

PR #878, opened 2020-04-22
Update design system for step: shared/relationship

Davidslv:
This is a shared partial between the following:
- applicant/relationship
- other_party/relationship
- respondent/relationship

image

Story: https://mojdigital.teamwork.com/#/tasks/20230253

### DIFF
--- a/app/views/steps/shared/_relationship_step.html.erb
+++ b/app/views/steps/shared/_relationship_step.html.erb
@@ -1,17 +1,13 @@
 <% heading = t('.heading', person_name: form_object.record.person.full_name, child_name:  form_object.record.minor.full_name) %>
 
-<h1 class="heading-xlarge gv-u-heading-xxlarge">
-  <%=  heading %>
-</h1>
-
-<%= step_form form_object do |f| %>
-  <%=
-    f.radio_button_fieldset :relation, { legend_options: { class: 'visually-hidden', text: heading } } do |fieldset|
-      fieldset.radio_input Relation::MOTHER
-      fieldset.radio_input Relation::FATHER
-      fieldset.radio_input(Relation::OTHER) { f.text_field :relation_other_value }
-    end
-  %>
+<%= step_form form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+  <%= f.govuk_radio_buttons_fieldset :relation, legend: { text: heading } do %>
+    <%= f.govuk_radio_button :relation, Relation::MOTHER, link_errors: true %>
+    <%= f.govuk_radio_button :relation, Relation::FATHER %>
+    <%= f.govuk_radio_button :relation, Relation::OTHER do %>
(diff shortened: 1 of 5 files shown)


app/views/steps/shared/_relationship_step.html.erb
```diff
@@ -1,17 +1,15 @@
...
-<%= step_form form_object do |f| %>
-  <%=
-    f.radio_button_fieldset :relation, { legend_options: { class: 'visually-hidden', text: heading } } do |fieldset|
-      fieldset.radio_input Relation::MOTHER
-      fieldset.radio_input Relation::FATHER
-      fieldset.radio_input(Relation::OTHER) { f.text_field :relation_other_value }
-    end
-  %>
+<%= step_form form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+  <%= f.govuk_radio_buttons_fieldset :relation do %>
```
zheileman, line 6: Hey I think we can inject the heading into the fieldset legend, like this:

```ruby
f.govuk_radio_buttons_fieldset :relation, legend: { text: heading } do ...
```

Give it a try. That way we don't need the separate `<h1>` tag.

app/views/steps/shared/_relationship_step.html.erb
```diff
@@ -1,17 +1,15 @@
...
+    <%= f.govuk_radio_button :relation, Relation::MOTHER, link_errors: true %>
```
zheileman, line 7: Only the first radio needs the `link_errors` flag, the others can be removed as don't have any effect.

app/views/steps/shared/_relationship_step.html.erb
```diff
@@ -1,17 +1,15 @@
...
+    <%= f.govuk_radio_button :relation, Relation::FATHER, link_errors: true %>
+    <%= f.govuk_radio_button :relation, Relation::OTHER do %>
+      <%= f.govuk_text_field :relation_other_value, label: { size: 'm' }, link_errors: true %>
```
zheileman, line 10: I think we've been using small size `label: { size: 's' }` for revealing elements (is that the default maybe? try removing it, if it outputs the small size, then we can go without it).

app/views/steps/shared/_relationship_step.html.erb
Davidslv: I was under the impression from looking at the validation on staging (for comparison) that it needs to validate the input of the text field when selecting `OTHER` since it throws a validation error without nothing on the text field?

zheileman: Text fields are always linked automatically as they have an unique ID so don't need that.

Radio fieldsets or Checkbox fieldsets produce an error on the whole fieldset, so in order to auto-focus and link properly to the radio or checkbox we want, we need to mark it with `link_errors: true`

Davidslv:
All addressed, rebased branch.

PR #893, opened 2020-04-29
Update other parties screens

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/20230250

Screenshot 2020-04-29 at 13 48 32
Screenshot 2020-04-29 at 13 48 24

### DIFF
--- a/app/views/steps/other_party/address_details/edit.html.erb
+++ b/app/views/steps/other_party/address_details/edit.html.erb
@@ -1,19 +1,22 @@
 <% title t('.page_title') %>
+<% step_header %>
 
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <%= step_header %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= govuk_error_summary %>
 
-    <h1 class="heading-xlarge gv-u-heading-xxlarge"><%=t '.heading', name: @form_object.record.full_name %></h1>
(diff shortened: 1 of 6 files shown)


app/views/steps/other_party/address_details/edit.html.erb
```diff
@@ -1,19 +1,22 @@
...
-      <%= f.text_field :postcode %>
+    <%= step_form @form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_text_field :address_line_1 %>
+      <%= f.govuk_text_field :address_line_2, label: { hidden: true } %>
+      <%= f.govuk_text_field :town, width: 'one-half' %>
+      <%= f.govuk_text_field :country, width: 'one-half' %>
+      <%= f.govuk_text_field :postcode, width: 'one-third' %>
 
-      <%= f.single_check_box :address_unknown, class: 'dont-know' %>
+      <%= f.govuk_check_boxes_fieldset :residence_requirement_met, legend: { tag: 'span', size: 'm', hidden: true } do %>
```
zheileman, line 17: I think this is not working as expected. The fieldset name should be also `:address_unknown` (like the checkbox) and then inspecting the html of the page, we need to ensure it is really creating a legend but with the hidden class. If not, then it is missing the Locale.

This is just for screen readers, the locale can set to "Current address unknown" same as we did with the date of birth unknown. It is not intended to visually show on the page.

I believe same problem exists in Respondents.

PR #896, opened 2020-04-30
Update Applicant Under Age

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/20340358

Follow up PR will remove respondent under age logic

image

### DIFF
--- a/app/views/steps/applicant/under_age/edit.html.erb
+++ b/app/views/steps/applicant/under_age/edit.html.erb
@@ -1,37 +1,40 @@
 <% title t('.page_title') %>
+<% step_header %>
 
-<div class="grid-row">
-  <div class="column-two-thirds">
-    <%= step_header %>
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl"><%=t '.heading' %></h1>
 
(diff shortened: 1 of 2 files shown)


app/views/steps/applicant/under_age/edit.html.erb
```diff
@@ -1,37 +1,40 @@
...
+          class: 'govuk-link'
         )
       %>
       <%= t('.court_and_tribunal_html', court_name: @court.name, court_and_tribunal_finder_url: court_and_tribunal_finder_url) %>
    </p>
 
-    <%= step_form @form_object do |f| %>
-      <%= f.single_check_box :under_age %>
+    <%= step_form @form_object, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |f| %>
+      <%= f.govuk_check_boxes_fieldset :under_age, small: true, legend: { tag: 'span', size: 's', hidden: true } do %>
```
zheileman, line 35: I think this check box will be best to not have the `small: true` as it is the main (and mandatory) action the user can take here and should appear big and prominent, not small.

In addition to that, check the hidden legend, I think is missing.

We can go with a short legend kind of like... "Litigation friend representation", the bulk is already in the checkbox itself.

PR #910, opened 2020-05-04
Update design system for: backoffice/audit

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/20230241
Screenshot 2020-05-04 at 16 13 43

### DIFF
--- a/app/views/backoffice/audit/_record.en.html.erb
+++ b/app/views/backoffice/audit/_record.en.html.erb
@@ -1,19 +1,19 @@
-<tr>
-  <td valign="middle" width="220"><%= record.created_at %></td>
-  <td valign="middle" width="200"><%= record.action.upcase %></td>
-  <td valign="middle"><%= record.author %></td>
+<tr class="govuk-table__row">
+  <td class="govuk-table__cell"><%= record.created_at %></td>
+  <td class="govuk-table__cell"><%= record.action.upcase %></td>
+  <td class="govuk-table__cell"><%= record.author %></td>
 </tr>
 
 <% if record.details.present? %>
-  <tr>
-    <td valign="middle" colspan="3" class="xsmall no-border">
-      <details data-block-name="audit-details">
-        <summary>
-          <span class="summary">
+  <tr class="govuk-table__row">
+    <td class="govuk-table__cell xsmall no-border" colspan="3">
+      <details class="govuk-details govuk-!-margin-bottom-0" data-module="govuk-details">
+        <summary class="govuk-details__summary">
+          <span class="govuk-details__summary-text">
             show details
           </span>
         </summary>
-        <div class="panel panel-border-narrow">
+        <div class="govuk-details__text">
           <%= record.details %>
         </div>
       </details>
--- a/app/views/backoffice/audit/index.en.html.erb
+++ b/app/views/backoffice/audit/index.en.html.erb
@@ -1,28 +1,26 @@
 <% title 'Back office: audit records' %>
 
-<div class="grid-row">
-  <div class="column-full">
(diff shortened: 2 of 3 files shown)


app/views/backoffice/audit/index.en.html.erb
```diff
@@ -1,27 +1,25 @@
...
       </thead>
 
-      <tbody>
+      <tbody class="govuk-table__body">
         <% if @report.any? %>
           <%= render partial: 'record', collection: @report, as: :record %>
         <% else %>
-          <tr>
+          <tr class="govuk-table__row">
             <td colspan="3">No records were found in the time period</td>
```
zheileman, line 23: td is missing the `class="govuk-table__cell"`

app/views/backoffice/audit/index.en.html.erb
```diff
@@ -1,27 +1,25 @@
...
-    <table class="backoffice-table">
-      <thead>
-        <tr>
-          <th>Timestamp</th>
-          <th>Action</th>
-          <th>Author</th>
+    <table class="govuk-table">
+      <thead class="govuk-table__head">
+        <tr class="govuk-table__row">
+          <th class="govuk-table__header">Timestamp</th>
```
zheileman, line 12: We can reduce the column widths using `govuk-!-width-one-third` (as the idea is to leave the 3rd column more space as it will contain full names and emails that are longer, so it doesn't wrap too much)

You can see before we were using a hardcoded width in the <td>

app/views/backoffice/audit/index.en.html.erb
```diff
@@ -1,27 +1,25 @@
...
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-full">
+    <h1 class="govuk-heading-xl">Back office: audit records</h1>
 
-    <p class="heading-medium">
-      Actions performed in the last 30 days
-    </p>
+    <h3 class="govuk-heading-m"> Actions performed in the last 30 days</h3>
 
-    <table class="backoffice-table">
```
zheileman, line 11: We still need to maintain some of the custom styling. Migrating where possible to new classes but if not possible, as with the check your answer, we use custom styles:
app/assets/stylesheets/local/app/backoffice.scss

app/views/backoffice/audit/_record.en.html.erb
```diff
@@ -1,19 +1,19 @@
...
-  <td valign="middle"><%= record.author %></td>
+<tr class="govuk-table__row">
+  <td class="govuk-table__cell"><%= record.created_at %></td>
+  <td class="govuk-table__cell"><%= record.action.upcase %></td>
+  <td class="govuk-table__cell"><%= record.author %></td>
 </tr>
 
 <% if record.details.present? %>
-  <tr>
-    <td valign="middle" colspan="3" class="xsmall no-border">
```
zheileman, line 9: It needs the custom styling.

app/views/backoffice/audit/index.en.html.erb
Davidslv: Do we want to keep the data smaller on purpose? I don't see much difference other than font size, but I don't know the reason behind that decision

zheileman: The smaller size is for the show details content, but it is not that important. The main thing would be the no-border so there is no extra border below like this:

Screen Shot 2020-05-05 at 08 43 42

If there are many audit entries, that extra border add a lot of noise and clutter.

Davidslv: image

zheileman: 👍

PR #912, opened 2020-05-05
Update design system for: backoffice/dashboard

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/20230241

### Side by Side with Staging
Screenshot 2020-05-05 at 14 52 51

Screenshot 2020-05-05 at 09 54 00

### DIFF
--- a/app/views/backoffice/dashboard/_completion.en.html.erb
+++ b/app/views/backoffice/dashboard/_completion.en.html.erb
@@ -1,38 +1,38 @@
-<tr>
-  <td valign="middle" width="200"><%= completion.completed_at %></td>
-  <td valign="middle" width="160"><%= completion.reference_code %></td>
-  <td valign="middle"><%= completion.submission_type || 'N/A' %></td>
-  <td valign="middle"><%= completion.court %></td>
+<tr class="govuk-table__row">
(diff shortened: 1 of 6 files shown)


app/views/backoffice/dashboard/_completion.en.html.erb
```diff
@@ -1,57 +1,57 @@
...
-                <dt>Urgent hearing</dt>
-                <dd><%= completion.metadata['urgent_hearing'] || 'N/A' %></dd>
-                <dt>Without notice hearing</dt>
-                <dd><%= completion.metadata['without_notice'] || 'N/A' %></dd>
-              </dl>
+        <table class="govuk-table">
+          <tr class="govuk-table__row">
+            <td class="govuk-table__cell no-border">
+                <dl class="govuk-summary-list govuk-summary-list--no-border">
+                  <dt class="govuk-!-font-size-14">Application started at</dt>
```
zheileman, line 18: There is a lot of repetition of classes here. Perhaps better to target the dt and dd from the stylesheet instead?
Like we had before:
https://github.com/ministryofjustice/c100-application/blob/master/app/assets/stylesheets/local/app/backoffice.scss#L13

PR #916, opened 2020-05-06
Restore the flash alert in the back office

zheileman:
This was refactored for the main layout but was not done in the back office layout.

However not using the same partial as the main layout as that one is for errors, and this is more for giving information.

Fixed the back office user menu as it was also done with the main user menu.

Screen Shot 2020-05-06 at 12 34 04


zheileman:
@Davidslv Backporting some work that was done in the main layout, but I forgot to do it in the back office too.

PR #927, opened 2020-05-11
Fix couple of links without govuk-link class

Davidslv:
This is not a critical one as this setting (https://github.com/ministryofjustice/c100-application/commit/4854f2ae304442a6699f4bd734544653cf8b3693#diff-f859e2848b07324f808ce1a5ccd9fcbaR5) has been turned on but I though it should be addressed anyway.


zheileman:
Hi @Davidslv, no harm in adding the class, but there is no need as these links are in the generated PDF.
The PDF does not use the govuk design system. In fact it uses a different stylesheet.

Davidslv:
Ah, yep, you are right, I did not see the file extension was just hunting for left overs. closing.

PR #956, opened 2020-06-01
Feature tests for Children Safety Concers with sub forms

Davidslv:
This commit enhances the feature test created during https://github.com/ministryofjustice/c100-application/pull/954
Supporting the sub forms during the journey, and stops at the Applicant Safety Concerns (to be covered in a future commit)

### DIFF
--- a/features/step_definitions/forms.rb
+++ b/features/step_definitions/forms.rb
@@ -12,6 +12,16 @@ When(/^I click the radio button "([^"]*)"$/) do |text|
   find('label', text: text).click
 end
 
+Then(/^I click "([^"]*)" for the radio button "([^"]*)"$/) do |text, legend|
+  find(
+    'legend > span', text: legend
+  ).ancestor(
+    'fieldset', order: :reverse, match: :first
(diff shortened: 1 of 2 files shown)


features/step_definitions/forms.rb
```diff
@@ -12,6 +12,12 @@
   find('label', text: text).click
 end
 
+Then(/^I click "([^"]*)" for the radio button "([^"]*)"$/) do |text, label|
+  node = find('.govuk-fieldset__heading', text: label)
```
zheileman, line 16: How about this. I think is more readable and we don't need to use class names as these can be fragile:

```ruby
Then(/^I click "([^"]*)" for the radio button "([^"]*)"$/) do |text, legend|
  find(
    'legend > span', text: legend
  ).ancestor(
    'fieldset', order: :reverse, match: :first
  ).find(
    'label', text: text
  ).click
end

Davidslv: That's nice, I'm going to give it a try 👍

PR #961, opened 2020-06-05
Validates that payment and submission types are fulfilled

Davidslv:
Story: https://mojdigital.teamwork.com/#/tasks/20765761

image

### DIFF
--- a/app/validators/application_fulfilment_validator.rb
+++ b/app/validators/application_fulfilment_validator.rb
@@ -19,6 +19,8 @@ class ApplicationFulfilmentValidator < ActiveModel::Validator
       ->(record) { [:children, :blank, edit_steps_children_names_path(id: '')] unless record.children.any? },
       ->(record) { [:applicants, :blank, edit_steps_applicant_names_path(id: '')] unless record.applicants.any? },
       ->(record) { [:respondents, :blank, edit_steps_respondent_names_path(id: '')] unless record.respondents.any? },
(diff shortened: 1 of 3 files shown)


app/validators/application_fulfilment_validator.rb
```diff
@@ -19,6 +19,8 @@ def validations
       ->(record) { [:children, :blank, edit_steps_children_names_path(id: '')] unless record.children.any? },
       ->(record) { [:applicants, :blank, edit_steps_applicant_names_path(id: '')] unless record.applicants.any? },
       ->(record) { [:respondents, :blank, edit_steps_respondent_names_path(id: '')] unless record.respondents.any? },
+      ->(record) { [:payment_type, :blank, edit_steps_application_payment_path] unless record.payment_type.present? },
```
zheileman, line 22: Given the submission step comes before the payment step (after we did the swap in the previous PR), I think it makes more sense to also change the order here.

spec/validators/application_fulfilment_validator_spec.rb
```diff
@@ -10,13 +10,61 @@ module Test
...
+      respondents: respondents,
+      payment_type: payment_type,
+      submission_type: submission_type,
+    }
+  end
 
   let(:children)    { [Object] }
   let(:applicants)  { [Object] }
   let(:respondents) { [Object] }
+  let(:payment_type) { PaymentType::SELF_PAYMENT_CARD.to_s }
```
zheileman, line 26: I think is better not to couple these tests to any specifics, and not use value objects.
After all, we are only testing for presence, so anything will do.

PR #969, opened 2020-06-11
Update Court to receive gbs code

Davidslv:
The GBS code is necessary to facilitate HMCTS to know which court to make a credit payment.

Once the user selects their Postcode, a request is made to find the nearest Court, each court should have a GBS code, but currently not all have it, that means it can be nil (null) but the key should be present in all of them.

Story: https://mojdigital.teamwork.com/#/tasks/20809040

### DIFF
--- a/spec/models/court_spec.rb
+++ b/spec/models/court_spec.rb
@@ -9,6 +9,7 @@ describe Court do
       "slug" => 'court-slug',
       "email" => 'family@court',
       "address" => address,
+      "gbs" => 'X123',
     }
   }
 
@@ -31,6 +32,10 @@ describe Court do
 
     it 'sets the address' do
       expect(subject.address).to eq(address)
+    end
+
+    it 'sets the gbs code' do
+      expect(subject.gbs).to eq('X123')
     end
 
     context 'log and raise exception when attributes not found' do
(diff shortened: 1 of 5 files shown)


spec/models/court_spec.rb
```diff
@@ -20,6 +20,14 @@
     }
   }
 
+  let(:gbs) { { 'gbs' => 'X123' } }
```
zheileman, line 23: Now that we are using the same approach as the email, we can move the `gbs` key-value to the `let(:data)` and remove the allow_any_instance_of, in favour of testing the behaviour, in a similar context as the one we have for the email, `context 'email'`

spec/models/court_spec.rb
```diff
@@ -87,7 +99,12 @@
         end
 
         context 'the API returned emails' do
-          let(:api_response) { {'emails' => [{"description" => "applications", "address" => "applications@email"}]} }
+          let(:api_response) do
+            {
+              'emails' => [{"description" => "applications", "address" => "applications@email"}],
+              'gbs' => nil
```
zheileman, line 105: I think no need to add the gbs code here, as this is testing only the email response.

spec/models/court_spec.rb
```diff
@@ -33,6 +41,10 @@
       expect(subject.address).to eq(address)
     end
 
+    it 'sets the gbs code' do
```
zheileman, line 44: As mentioned, this can be moved to its own context, like we do with `email`

PR #1011, opened 2020-07-20
Add consent order text for online and print and post

njseeto:
We are still waiting for instructions on where to place the confirmation content on the 'digital submission' page.

After changes, it will look like this:

Screenshot 2020-07-21 at 11 39 42

Story: https://mojdigital.teamwork.com/#/tasks/21212954

### DIFF
--- a/spec/views/steps/completion/confirmation/show.html.erb_spec.rb
+++ b/spec/views/steps/completion/confirmation/show.html.erb_spec.rb
@@ -1,0 +1,67 @@
+require 'rails_helper'
+
+RSpec.describe 'steps/completion/confirmation/show', type: :view do
+  let(:consent_order) { GenericYesNo::YES }
+  let(:under_age) { true }
+  let(:applicants_collection_double) { double(under_age?: under_age) }
+  let(:c100_application) { C100Application.new(id: '449362af-0bc3-4953-82a7-1363d479b876', created_at: Time.at(0), consent_order: consent_order) }
+
+  before do
+    allow(c100_application).to receive(:applicants).and_return(applicants_collection_double)
+
+    assign(:c100_application, c100_application)
+
+    assign(
+      :court,
+      double('Court',
+        name: 'Test court',
+        email: 'court@example.com',
+        slug: 'test-court'
+      )
+    )
+
+    render
+  end
+
+  context 'when applicant is under age' do
+    it 'should render the `under_age_section` partial' do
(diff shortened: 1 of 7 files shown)


spec/views/steps/completion/confirmation/show.html.erb_spec.rb
```diff
@@ -0,0 +1,67 @@
+require 'rails_helper'
+
+RSpec.describe 'steps/completion/confirmation/show', type: :view do
+  let(:consent_order) { GenericYesNo::YES }
+  let(:under_age) { true }
+  let(:applicants_collection_double) { double(under_age?: under_age) }
+  let(:c100_application) { C100Application.create(consent_order: consent_order) }
```
zheileman, line 7: Quick suggestion, here instead of creating a record, you can just instantiate one, but the trick for it to work (as you've probably seen it will fail) is to instantiate it with the `id` and the `created_at` attributes already set, like this (the ID can be anything conforming to an UUID format):

```ruby
C100Application.new(
  id: '449362af-0bc3-4953-82a7-1363d479b876',
  created_at: Time.at(0),
  consent_order: consent_order
)
```

The ID and the create_at are used for the reference code. Another alternative is to still instantiate a record instead of create, but no need to pass the id and the created_at if you mock the `reference_code`:

```ruby
allow(c100_application).to receive(:reference_code).and_return('reference-code')
```

Both options are equally valid and will save a hit in the database.

spec/views/steps/completion/what_next/show.html.erb_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'rails_helper'
+
+RSpec.describe 'steps/completion/what_next/show', type: :view do
+  let(:consent_order) { GenericYesNo::YES }
+  let(:c100_application) { C100Application.create(consent_order: consent_order) }
```
zheileman, line 5: Same as above.

spec/views/steps/completion/confirmation/show.html.erb_spec.rb
njseeto: Ah nice one! We tried instantiating a new one but didn't add those attributes, so like you mentioned it failed.

Agreed, instantiating one will be nicer than creating so will make the change on both this occasion and in the below `what_next/show.html.erb.spec.rb` file. Thanks!

PR #1019, opened 2020-07-27
Remove dependency on `email_validator` gem

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21318405

The code in the gem is minimum and we can move it to our service and simplify it, reducing an external dependency and minimising the risk of this dependency being updated/changed and then our validations could break.

In fact this gem was already locked in version `< 2.0.0` because in greater version they introduced a change that didn't fit well with our validations (they made the regex much more simple thus emails like `a@b` would be valid).

I've moved the code and cleaned up to remove anything we don't need so it is just the regex.

Also as part of this I've added a very simple list of common domain typos we see frequently to try to avoid some of them. This was not part of the original gem.

For reference this is the code of the validator in the gem (< 2.0.0) I've based this work on:

https://github.com/K-and-R/email_validator/blob/ee5a44dcb936e9dae4a179e55c8be85742b601c3/lib/email_validator.rb

### DIFF
--- a/app/validators/email_validator.rb
+++ b/app/validators/email_validator.rb
@@ -1,0 +1,32 @@
+class EmailValidator < ActiveModel::EachValidator
+  EMAIL_REGEX = /\A\s*([-\p{L}\d+._]{1,64})@((?:[-\p{L}\d]+\.)+?\p{L}{2,})\s*\z/i
+  COMMON_DOMAIN_TYPOS = %w[
+    gamil.com
+    gmail.con
+    gmail.co
+    hotmial.com
+    hotmail.con
+    hotmail.co
+  ].freeze
+
+  def validate_each(record, attribute, value)
+    if valid_format?(value)
+      record.errors.add(attribute, :typo) if domain_typo?(value)
+    else
+      record.errors.add(attribute, :invalid)
+    end
+  end
+
+  private
+
+  def valid_format?(value)
+    EMAIL_REGEX.match?(value)
+  end
+
+  # Very simplistic but we see these typos quite frequently
+  def domain_typo?(value)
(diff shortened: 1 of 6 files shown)


app/validators/email_validator.rb
```diff
@@ -0,0 +1,32 @@
+class EmailValidator < ActiveModel::EachValidator
+  EMAIL_REGEX = /\A\s*([-\p{L}\d+._]{1,64})@((?:[-\p{L}\d]+\.)+?\p{L}{2,})\s*\z/i
+  COMMON_DOMAIN_TYPOS = %w[
```
Davidslv, line 3: 👍

config/locales/errors/en.yml
```diff
@@ -2,6 +2,7 @@ en:
   dictionary:
     invalid_email_error: &invalid_email_error "Enter an email address in the correct format, like name@example.com"
     blank_email_error: &blank_email_error "Enter an email address"
+    email_typo_error: &email_typo_error "The email address appears to have a typo"
```
Davidslv, line 5: Some optional phrases:

- The email address appears to be mistakenly typed
- The email address seems to be mistaken
- The email address seems to be incorrect

zheileman: Yeah! I agree, `typo` sounds a bit slang.

I've just done some google searches (yeah I know, very scientific 🤣) and the combinations that gets more results are "seems to have" (as opposite to "appears to have") an "typing error" (as opposite to "typing mistake" or "mistakenly typed").

So I think we can go with: "The email address seems to have a typing mistake".

Davidslv: Sounds good 👍

PR #1020, opened 2020-07-28
Bump rubocop and cleanup configuration

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21333703

We were using a very old version, mostly inherited from a previous project.
This removes the version lock on rubocop and let it update to the latest possible version (but still not the latest available one due to another dependency, that I will deal with it in a separate PR)

I've done a big cleanup of the rules we had in the config file to only leave the ones we want for this project as we've been using some code format consistently.


zheileman:
Note there is still a dependency stopping us from bumping to the latest version (0.88.0).
I will deal with that in a separate PR to continue bumping the whole project.

njseeto:
LGTM! 👍

PR #1051, opened 2020-08-19
Add permission sought and details to main journey

zheileman:
Follow-up to PRs #1048 and #1049.

Incorporate these two questions in the main journey, right after finalising the international sub-journey.

For now, it will always show, as we are not checking the permission rules. This will be done in the next PR.


njseeto (approved):
great! 🚀

PR #1052, opened 2020-08-20
Do not touch `updated_at` on navigation updates

zheileman:
This was messing a bit with some reports and statistics we have, because each time there is a navigation change (`navigation_stack` gets updated) it did a save and touched the `updated_at` magic column too.
And we tend to sort many of these report by updated_at.

The timestamp changes separately from the `navigation_stack` each time there is an attribute update, i.e. when the user is answering questions and entering details.


zheileman:
Ok that's annoying. It works on local but fails on CI. Will try to find the reason.

njseeto (approved):
Looks good 💯

PR #1053, opened 2020-08-20
Implement permission to apply rules

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21392567
Parent ticket: https://mojdigital.teamwork.com/#/tasks/21392255

This _looks_ like a big PR but most of it was moving and renaming an existing class/spec to a different namespace.
Skip to commit b57b713 for the important bits.

This rules class will be in charge of deciding if an application (as a whole) requires permission to apply or not (non-parents).

It is slighly more complex than the other class because here we need to run through all the relationships applicant-child and based on the answers to the 9 questions AND the orders selected for the child in the relationship, decide if they have the permission or not.


njseeto (approved):
LGTM 🚀

PR #1059, opened 2020-08-24
Extend CYA permission tests

zheileman:
Just a minor refactor in line with the PDF tests, so it covers all possible options (including the permission question not being asked and thus being `nil`).


njseeto (approved):
Looks great! Nice clean up ✨

PR #1060, opened 2020-08-25
Log in the audit table permission info

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21633583

This will log if the application had any children with a special guardianship order in force, and if permission was sought (or is not required, which happens if the question is not asked and thus attribute is `nil`).
It will also show this info in the back office.

Note: Even if we've not released yet non-parents, these are not breaking changes and can be released at any time.


Davidslv (approved):
👍

PR #1062, opened 2020-08-26
Split international/litigation PDF question

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21483874

Split the question that previously was "Case with an international element or factors affecting litigation capacity?" into 2 separate questions, following changes that will be introduced as well in the paper form.

### Before
Screen Shot 2020-08-26 at 10 28 50

### After
Screen Shot 2020-08-26 at 10 29 29


njseeto (approved):
great! 🙌

PR #1063, opened 2020-08-26
Cleanup no longer used email consent in screener

zheileman:

Note: although it looks big, it is just removing stuff :smile:

As part of the soon to be released non-parents, and removal of the parent screener question, we can also get rid of this other question that is currently hidden, because we intend to remove the screener completely soon after non-parents.

If we ever need to gather emails for research purposes, we will need to think of a better way, probably embedded as part of the main application, and not in a screener.

This PR can be merged and deployed without risk. Follow-up PR to this one, the removal of the parent question, can't be merged just yet.


Davidslv (approved):
🚜

PR #1064, opened 2020-08-27
Remove parent question from screener

zheileman:
Parent ticket: https://mojdigital.teamwork.com/#/tasks/21392255

This question will no longer be necessary once we release the
non-parents work next week.

Cleanup of all code related to this question as well as the kickout
page (when answered No), routes, locales, views, etc.

To be merged and released to prod only when we enable non-parents.


njseeto (approved):
LGTM! 🚀

Davidslv (approved):

PR #1065, opened 2020-08-27
Tweak permission undecided rules

zheileman:
Having a child in the application with a special guardianship order in force, automatically makes the application require permission.

Before, we were still asking the 9-questions in these cases, but with this change, we don't need to, and we assume straight away the permission will be needed.

Simplified so the `#permission_needed?` is no longer needed and is more clear what is going on.


njseeto (approved):
nice! 🏅

Davidslv (approved):
awesome!

PR #1067, opened 2020-09-01
Minor copy tweaks for non-parents

zheileman:

Reduce the places where we assume the applicant is a parent of the children (as we are about to release non-parents).


njseeto (approved):
Looks good 👍

PR #1070, opened 2020-09-02
Temporarily disable online payments to Preston court

zheileman:
GOV.UK Pay has a length limit in the metadata values of 50 characters.

We use the metadata to store the name of the court where the payment was made. Unfortunately we didn't anticipate and some courts have a lengthy name, like this one, taking us over the 50 limit and getting an error when trying to create an online payment.

The name of this court is `Preston Crown Court and Family Court (Sessions House)` which is 53 chars but I've seen other courts with even 64 chars.

Temporarily disable this court. I've contacted GOV.UK Pay to raise the limit in the meantime and will be re-enabled as soon as possible.


Davidslv (approved):
😢

PR #1072, opened 2020-09-07
Add to audit table the applicant(s) relation to child(ren)

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21737933

This is in relation to the difficulty we have to know how many fathers, mothers, etc. triggered non-parent applications (those with the `permission_sought` attribute not `nil`).

If we store an array of the relations (for example `["father", "other"]`) at least we can map this in a chart although still we will not be certain what was the trigger of the permission (but chances are 99% will be the "other" relationship).

There is a rake task to be run manually once deployed, to retroactively populate existing non-parents completed applications (there are a few, less than 20).
New ones will contain this metadata detail by default.


njseeto (approved):
Great! 👍

Davidslv (approved):
👍

PR #1073, opened 2020-09-08
Rollout of online payments batch 2 courts

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21709675

**Online payments rollout -- batch 2**

Instead of having to add one by one all the courts that accept online payments (as these are a lot after batch 2), a blocklist is more efficient and cleaner, adding only those courts that for whatever reason can't take online payments.

Also, disallow online payments for courts without a GBS code set.

Currently we are adding the courts from batch 3 as these courts will go live next week, as well as a few courts with names taking us over the 50 chars limit on GOV.UK Pay.

Once Pay increase the limit we can removed those. And next week we will be able as well to hopefully remove the remaining ones, leaving the blocklist empty (that's the theory).

### DIFF
--- a/app/models/court.rb
+++ b/app/models/court.rb
@@ -1,4 +1,6 @@
 class Court
+  UNKNOWN_GBS = 'unknown'.freeze
+
   attr_reader :name, :slug, :email, :address, :gbs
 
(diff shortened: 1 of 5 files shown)


Davidslv (approved):
👍

PR #1075, opened 2020-09-08
Disable Newport court from online payments

zheileman:
As we did with other courts, this one has a name over 50 chars, and unfortunetely GOV.UK Pay have not released the increase yet, so attempts to call the API with this court name will fail with code P0102.

Add to the sentry exception the application ID if we have it to help debug wich courts are failing more easily.


njseeto (approved):
LGTM! 👍

PR #1076, opened 2020-09-09
Remove financial abuse wording

njseeto:
Do not merge until we have written confirmation for this change.

Task: https://mojdigital.teamwork.com/#/tasks/21964062

#### Before
old

#### After
new

#### Before
old

#### After
new

#### Before
old 2

#### After
new 2


zheileman:
There are a couple more places where we mention this that I imagine must be removed also.
In the financial abuse page for applicant and for children.

Like this:
Screenshot 2020-09-09 at 12 49 29

zheileman (approved):
👍 thanks!

PR #1077, opened 2020-09-09
Reduce level of payment debug captured in Sentry

zheileman:
We had this code to report payment failures (as in, users cancelling their payment, or payment card rejected) for some time while we made sure online payments were working fine.

We can now remove this info logs.

Any exception produced by the payments API will regardless be reported  to Sentry as those are important errors.

We can also see all these details and errors in the Payments transaction list.


njseeto (approved):
nice! 🚀

PR #1078, opened 2020-09-14
Enable Pay in courts disabled due to char limit

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21709675

We disabled temporarily these 5 courts as their names were over the 50 chars metadata limit.

Now Pay team have confirmed they've increased this limit:

```Custom metadata on payments can now have values that are strings of up to 100 characters (previous limit was 50)```

### DIFF
--- a/config/govuk_pay.yml
+++ b/config/govuk_pay.yml
@@ -12,15 +12,6 @@ blocklist:
   - barrow-in-furness-county-court-and-family-court
   - liverpool-civil-and-family-court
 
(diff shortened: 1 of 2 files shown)


zheileman:
Note: this PR will fail while CTF is still down due to the cuke integration tests.
Once CTF is up and running we can re-run the job to pass the tests. And hopefully merge and release (tomorrow).

PR #1079, opened 2020-09-15
Prepare for the screener refactor

zheileman:

This PR may be a bit big, I've split into smaller commits with more details

This in itself does pretty much nothing other than prepare the path for the next PRs (not sure how many) in the coming days in order to slowly refactor the current `screener_answers` into a separate `Courts` table, and decouple the children postcode and this court data organising it in a better way in the database.

If all goes well the idea is to have a table of `Courts` (one for each court we have) that is auto-populated as we "see" that court (we don't need to pre-populate this table ourselves) and then, use the `slug` of the court as a primary key ID and assign this to each application that goes to that same court. This way we denormalise the data (1 court can have many applications) instead of duplicating again and again the court data in each of the applications created (current table `screener_answers`, field `local_court`).

This first PR does not affect (or should not) users, and nothing is visible, everything is behind the scenes.
The new table `courts` is not yet being used. That will come in the next PR.


Davidslv (approved):
👍 nice work!

PR #1083, opened 2020-09-21
Reduce race-condition edge cases in online payments

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21915640

We've had a couple of cases (out of almost 500 successful transactions, so really edge case) where a payment was created on GOV.UK Pay twice at the (almost) same time.
**Note: payment in this context is the "attempt or intent", not the actual take of funds.**

One of these transactions usually will progress to success and funds will be taken, the other will remain "In progress" for up to 90 minutes and then time out.

Problem with this is, in our database we need to link the c100 application to the corresponding online payment, with a unique ID that Pay provides to us after the creation of each payment.

As 2 payments are created sequentially almost at the same time, only one of these transactions will "win the race" and write its unique payment ID to our database.
This usually is not a problem, if the second transaction is the one that we store in the database and the one that completes the payment online. But if it is the first instead, then we will not be able to know if the payment has been successful or not (as we have a different ID).

This PR tries a first attempt (hopefully it is enough) to reduce as much as possible the points where a race-condition can happen, as well as to reuse payment intents where possible in order to continue an "in progress" payment and to not create a new unnecessary payment.

**Individual commits with more context.**

### DIFF
--- a/app/models/payment_intent.rb
+++ b/app/models/payment_intent.rb
@@ -12,6 +12,10 @@ class PaymentIntent < ApplicationRecord
     update_column(:nonce, nil)
   end
 
+  def in_progress?
+    state['finished'].equal?(false)
+  end
+
   private
 
   def init_nonce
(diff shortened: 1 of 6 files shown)


Davidslv (approved):
👍

PR #1084, opened 2020-09-22
Read court details from the new table

zheileman:

We will expose a `#court` method in the C100Application model, that will check if we have a court record in the new `courts` table, and if not, fallback to the legacy `screener_answers` table.

We can remove this method once we don't have any more records using the old table.

Changed a lot of the `screener_answers_court` references. Some remain (specifically in the developer tools bypass functionality) that will be refactored in the next PR.


njseeto (approved):
LGTM! 👍

PR #1085, opened 2020-09-22
Refactor bypass functionality

zheileman:
Follow-up to PR #1084.

The bypass is not really a "bypass screener" anymore, but instead a "bypass postcode" as that is the only thing we have left of the screener.

Also, do not write to the old `screener_answers` table, and instead we use the new `courts` table mechanism.


njseeto (approved):
Looks great! 🚀

PR #1086, opened 2020-09-22
Refactor `children_postcodes`

zheileman:

Similar to what we did in PR #1084, we now want to start using only the new `children_postcode` DB field (which is part of the C100Application record) instead of the previous `children_postcodes` (note plural) in the old ScreenerAnswers record.

We need to maintain backwards compatibility until all applications use the new attribute.


njseeto (approved):
Nice! ✨

PR #1087, opened 2020-09-23
Rename namespace from screener to opening

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21934206

It looks like a huge PR but it is just a noisy renaming. Functionality is still the same, only URLs have changed.

This is in preparation for moving the postcode into the main journey and removing the rest of the screener pages.

The postcode step will become part of an `Opening` group of questions where we will also include the first 2 questions of the journey, at the moment:

* `/steps/miam/consent_order`
* `/steps/miam/child_protection_cases`

So it will become a micro-journey, kind of a screener but not really a screener, thus the rename to `Opening` as per service designer consultation.

It will act like our current screener, meaning you must complete all 3 questions in order to progress further or be able to jump to any other part of the form, including the check your answers.


njseeto (approved):
Looks good! 💯

PR #1088, opened 2020-09-24
Remove screener and move postcode into main journey

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/21934206

Follow-up to PR #1087.

**Warning: gigantic PR but mostly due to moving/renaming things, as no new code has been introduced. If anything, some code has been deleted.**

Individual commits for more context although some of those commits are quite big in itself.

This is a step closer (we're almost there) to remove completely the screener. In fact, in this PR it is removed, although in a follow-up PR there are still some bits and pieces to do.

The homepage of the service has now become the page that came after the screener, and the postcode has moved to be the first question of the main journey. Also, the following 2 questions were previously on the MIAM namespace and that didn't make too much sense as those are not really related to the MIAM, so I've put these in the `Opening` group.

The rest is just moving around the steps, renaming and changing URLs and little more, but it is a very noisy PR 😢


njseeto (approved):
Looks great! 🚀

PR #1089, opened 2020-09-24
Update CYA to move 2 questions out of MIAM group

zheileman:
These are the 2 questions we moved in the previous PR #1088 out of the MIAM group as they are not directly linked to MIAM.

They will continue working the same but just outside the MIAM requirement group.
This has been agreed and discussed with the service designer.

Before:
Screenshot 2020-09-24 at 14 50 00

After:
Screenshot 2020-09-24 at 14 48 48

### DIFF
--- a/app/presenters/summary/html_presenter.rb
+++ b/app/presenters/summary/html_presenter.rb
@@ -8,6 +8,7 @@ module Summary
 
     def sections
       [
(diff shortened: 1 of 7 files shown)


njseeto (approved):
Nice 🙌

PR #1092, opened 2020-09-25
Improve email common typos validator

zheileman:
Gmail and Hotmail are the most commonly used email domains by our users, so by making sure these adhere to the short list of valid domains for each of these providers, we can reduce a lot of failures and typos without having to create a big blocklist.

Google only has 2 (public) email domains for users, so this one is easy.

Hotmail is no longer active but existing addresses continue to work so we need to support these.


njseeto (approved):
Looks good 🙌

PR #1095, opened 2020-10-01
Change order of the payments report in back office

zheileman:
Ordering by `updated_at` is more accurate, as a payment intent could have been created a few days ago but not completed the payment, and left to expire, but then user comes back and complete the payment (the created_at would remain the same but the updated_at will change).

This is just for the reports in the back office and do not affect anything else.


njseeto (approved):
👍

PR #1097, opened 2020-10-05
Add task to purge old back office audit records

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22069087

Similar to other automated daily tasks, create one to clean up older than 90 days back office audit records. These are of no use and they will continue growing otherwise.


Davidslv (approved):
👍

PR #1098, opened 2020-10-08
Add functionality to back office to complete applications

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22084915

This is a first step towards improving the situation where some online payments are left incompleted by mistake, as the user think the pre-auth is the payment, and then their application is not submitted to court.

For saved applications it is not much of an issue as the user can resume their application and retry the payment.
But not-saved applications can't be resumed and the user might have to start over again from the beginning which is not great.

With this new functionality exposed in the back office, it is now possible to search the reference of an incomplete application and if found, will give some details.

If the online payment failed or expired (and the application was not already completed by other means) then there is a button to complete it ad-hoc, changing the payment method to "phone".

This is mainly intended for users that can't resume the application as they didn't save it and contact us in the helpdesk mailbox, and hopefully avoid them having to start over again.

Some refinements might need to come up as a follow-up PR.

Screenshot 2020-10-08 at 12 14 41

### DIFF
--- a/app/controllers/backoffice/dashboard_controller.rb
+++ b/app/controllers/backoffice/dashboard_controller.rb
@@ -15,12 +15,48 @@ module Backoffice
 
       render :lookup && return if @form.invalid?
 
-      @report = CompletedApplicationsAudit.where(reference_code: @form.reference_code)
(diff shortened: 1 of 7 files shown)


Davidslv (approved):
🐑 🇮🇹

PR #1099, opened 2020-10-09
Cleanup list of slugs blocked for online payments

zheileman:
This was introduced as a temporary measure while existing applications in the database refreshed their GBS code.

There is no longer a need to explicitly block these, as all the existing applications (and future ones) will have the GBS code of these courts to `nil` or `unknown` so will be automatically blocked from taking online payments.


Davidslv (approved):
👍

PR #1101, opened 2020-10-14
Track more precisely CYA change links

zheileman:

This small tweak is related to the above ticket and will allow us to track more precisely which "change" links are more frequently clicked in the CYA page.

Instead of just having an event with a "change link" label, the label now will be a bit more specific, for example: `change link: miam_certification_details`.

This will then allow us to make an informed decision on what kind of improvements we can make to return the user back to the CYA after they've changed something.


njseeto (approved):
Looks great 🙌

PR #1102, opened 2020-10-15
Copy tweak in PDF miam certification section

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22085448

This is related to the MIAM certification rework that is undergoing, however is somehow unrelated and just a minor copy change to make it more clear some sections does not apply if the applicant is exempt from going to MIAM due to a consent order, or child protection cases.

Before it looked like the applicant had answered those questions with NO, however, that's not how it works, because if the applicant is exempt due to any of the first 2 questions (consent or child protection cases) then the rest of the miam certification questions are skipped, thus the applicant is not really answering with NO, is the system who is marking those as not required.

We use the same copy as we use in other places to mark not required sections, for consistency.

Also, removed the default value for these 2 initial questions because these are mandatory, can't be left unanswered so the default value is not needed.

### Before
Screenshot 2020-10-14 at 15 22 13

### After
Screenshot 2020-10-14 at 15 45 08

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -35,8 +35,9 @@ class SessionsController < ApplicationController
       court: find_or_initialize_court,
       status: 1,
       # fill out the first steps, so `save and return` button shows
-      consent_order: 'no',
-      child_protection_cases: 'no',
+      # if a value already exists we use it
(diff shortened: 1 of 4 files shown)


njseeto (approved):
LGTM! 🚀

PR #1104, opened 2020-10-15
Revert "Hotfix for Court Tribunal Finder invalid certificate"

zheileman:
This reverts commit b2ad1146f3c41538c6e7b9daef199b8f05e50b49.

Problems with CTF certificate seems to have been solved now. We didn't need to release this to PROD after all.


njseeto (approved):
Thanks for looking into the issue! 🙏

PR #1105, opened 2020-10-19
Refactor `update_navigation_stack` code into separate class

zheileman:

This is the first part of a "potential" improvement to the check your answers page. Mainly a refactor at this point.

Main idea is to set a flag in the session if we detect the user accessed a page through the `change link` in the CYA. This flag will later on (in a follow-up PR) allow us to decide if we can take the user back straight to CYA.

In this PR we set the session flag but do nothing.


Davidslv (approved):
nice 👍

PR #1106, opened 2020-10-20
Enable some CYA changes to fast-forward back

zheileman:

Follow-up to PR #1105.

This is the actual implementation of the fast-forward functionality, that started with the previous PR.

For now there is just a handful of steps that have this enabled, defined as regexp in `FAST_FORWARD_STEPS`.

Basically all the abuse concerns detail steps either for applicant or children, and the application details step.
This is mainly a proof of concept, so starting small and we will think how to scale up.

Exposed a helper method `#fast_forward_to_cya?` that can be used in the `#update_and_advance` to know if after updating a form object it can jump straight back to the CYA page, and is also used in the `#step_header` method, so we can modify the `back link` to point to the CYA as well.

I had fun with the navigation stack. Basically when using fast forward we can't really use it, otherwise if jumping from CYA to any intermediate step, will remove from the stack all the following steps from that intermediate point up until the CYA.

So I came to the conclusion that when using the fast-forward, we can let the navigation stack intact, so when returning to CYA everything is still there as if the user would have go one by one through all the intermediate steps again (pretty much as it happens with any other step not having fast-forward enabled).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,7 +30,7 @@ AbcSize:
   Max: 24
 
 CyclomaticComplexity:
-  Max: 15
+  Max: 12
 
 Documentation:
   Enabled: false
--- a/app/controllers/step_controller.rb
+++ b/app/controllers/step_controller.rb
@@ -11,13 +11,17 @@ class StepController < ApplicationController
   end
   helper_method :previous_step_path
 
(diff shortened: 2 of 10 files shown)


Davidslv (approved):
Thanks 👍

PR #1109, opened 2020-10-23
Add more steps to CYA fast-forward

zheileman:

Follow-up to previous PR #1106.

Add some more steps that can safely do a fast-forward to the CYA when editing something.

More to come in next PRs but I think for now these will cover a good chunk of the potential changes a user might need to do.

It will most likely be that we can enable the fast-forward for any step with ends with `_details` so eventually we will be able to simplify a lot the regex but for now it is easier to add one at a time.


Davidslv (approved):
👍

PR #1116, opened 2020-11-04
Bump some gem versions

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22398318

Only some minor releases except `wicked_pdf` but I've checked and there are no breaking changes or regressions as far as I can see.


Davidslv (approved):
Nice one

PR #1119, opened 2020-11-11
Ability to set a percentage to show/hide the research consent step

zheileman:
We implemented over a week ago the ability to gather users's consent for user research purposes, by showing them a page where they can enter their email address.

This was present some time ago but we removed it. Now we've reintroduce it, but we are getting quite a few too much candidates and will be difficult if not impossible to really manage all of them and the list is growing every day.

As a potential improvement, instead of turning off completely the gathering of this consent, with this PR we prepare the path for, if we want, show this consent only to a percentage of the new applications.

By default at the moment is 100% meaning all applications/users are shown the consent step, but we can let's say in a week decide to reduce this to 50%, and then the consent will be shown only to a 50% of the new applications, in a pesudo-random, deterministic way (based on the application creation date).

The little utility class to calculate the sampling can be reused of other purposes and a/b testing if we wanted to, as it is abstract and simple enough.

### DIFF
--- a/app/services/c100_app/opening_decision_tree.rb
+++ b/app/services/c100_app/opening_decision_tree.rb
@@ -29,10 +29,10 @@ module C100App
       if court
         c100_application.update!(court: court)
 
-        if Rails.configuration.x.opening.hide_research_consent_step
-          edit(:consent_order)
-        else
+        if show_research_consent?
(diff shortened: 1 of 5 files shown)


Davidslv (approved):
👍 nice!

PR #1120, opened 2020-11-11
Remove solicitor address text area from contact details step

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22442150

Follow-up to PR #1118.
In this PR we are removing the existing text area in the contact details step, as we've now introduced this same text area in a separate "address details" step.

Update the validations and the order of the fields in the contact details form object.

Update the solicitor decision tree to introduce this new address details step, before the contact details one.

Note: still pending, in follow-up PRs, update the CYA page to be able to get back to the new address step. Also enable the fast-track.

Once we have all this refactored and up and running, we can start structuring the text area into individual fields.

Screenshot 2020-11-11 at 15 17 16

Screenshot 2020-11-11 at 15 17 25

### DIFF
--- a/app/forms/steps/solicitor/contact_details_form.rb
+++ b/app/forms/steps/solicitor/contact_details_form.rb
@@ -5,14 +5,13 @@ module Steps
 
       has_one_association :solicitor
 
-      attribute :address, StrippedString
-      attribute :dx_number, StrippedString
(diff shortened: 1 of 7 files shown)


zheileman:
@Davidslv ping

Davidslv (approved):
👍

PR #1130, opened 2020-12-15
Add kick out page when there is no document from mediator

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22821248

As part of the changes proposed for the MIAM journey, a new kick out page needs to be added if the user answer NO to the question `Have you got a document signed by the mediator?`

Updated the decision tree and some copy.

Screenshot 2020-12-15 at 11 48 37


Davidslv (approved):
👍

PR #1131, opened 2020-12-16
Remove MIAM 4 months validation

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22823916

We've received some feedback and confirmation that the 4 months requirement does not apply to the MIAM (there is an exemption where the 4 months apply tho but not in the main journey).

We've had approval for the removal of the validation.

Still pending copy to be removed or reworded in a couple of places, in a follow-up PR when that is decided.


Davidslv (approved):
🚀

PR #1132, opened 2020-12-16
Create MIAM mediator exemption yes-no step and DB attr

zheileman:

This PR adds the new step (yes-no question) with the controller, form object and view, as well as the new DB attribute.

It is still not linked in the decision tree and not visible, unless it is accessed directly by going to the URL.

In following PRs it will be linked to the decision tree as well as adding the new question to the CYA and the PDF.

Screenshot 2020-12-16 at 13 33 03


Davidslv (approved):
👍

PR #1133, opened 2020-12-16
Update MIAM reset mechanism in multiple steps

zheileman:

Follow-up to PR #1132

Update the decision tree (but the new step is still not showing) so the journey is still the same.

Updated several form objects to perform the correct reset of attributes when changing from yes to no, or viceversa, in order to not leave inconsistent selected options across the journey.


Davidslv (approved):
👍

PR #1134, opened 2020-12-16
Add mediator’s exemption to CYA and PDF

zheileman:

Follow-up to PRs #1132 and #1133 

* Add the new question to the decision tree so it will show after answering NO to the question `Have you attended a MIAM?`

* Add the answer to this new question to the CYA and to the PDF.


Davidslv (approved):
👍

PR #1135, opened 2020-12-17
Show page to find a mediator when user has not attended MIAM

Davidslv:
As part of the changes proposed for the MIAM journey, we are closing the loophole where a user could continue with the application even without exemption and no attendance to MIAM.
This is in line with the paper form.

If the applicant didn't attend a MIAM, didn't have a meditor's exemption and didn't have another reason not to attend, then instead of continuing to safety concerns like now, (/steps/safety_questions/start) show a kick out page as per design.

Screenshot 2020-12-16 at 15 56 40

I've confirmed that there was no need to change the copy of the existing page.

Screenshot 2020-12-17 at 14 10 05

Story: https://mojdigital.teamwork.com/#/tasks/22821224

### DIFF
--- a/app/services/c100_app/miam_decision_tree.rb
+++ b/app/services/c100_app/miam_decision_tree.rb
@@ -29,7 +29,7 @@ module C100App
       if question(:miam_exemption_claim).yes?
         edit('/steps/miam_exemptions/domestic')
       else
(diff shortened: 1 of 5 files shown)


zheileman (approved):
👍

PR #1136, opened 2020-12-17
Copy changes to remove references to 4 months MIAM

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22823916

Copy changes proposed by content designer as a follow-up to PR #1131 where we removed the 4-months validation.


Davidslv (approved):
👍

PR #1137, opened 2020-12-18
Reintroduce MIAM exemptions playback and refactor

zheileman:
Ticket: https://mojdigital.teamwork.com/#/tasks/22821224

After the applicant goes through the list of MIAM exemptions, if they don't select any, we show them the kick out page (that was done in PR #1135) but if the applicant select one or more exemptions we should show them the playback page.

This playback page used to be after the safety concerns journey but because we now don't let the user continue to this journey we need to show it before.

As a consequence a bit of refactor and simplification in other decisions tree have been made as the logic is now much more simpler.

Screenshot 2020-12-18 at 11 26 56

### DIFF
--- a/app/services/c100_app/abuse_concerns_decision_tree.rb
+++ b/app/services/c100_app/abuse_concerns_decision_tree.rb
@@ -7,7 +7,7 @@ module C100App
       when :details
         after_details_step
       when :contact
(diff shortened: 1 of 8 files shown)


Davidslv (approved):
Looks good to me 👍

PR #1141, opened 2021-01-07
Disable research consent step

zheileman:
We had this disabled via ENV variable in production environment (value 0) but the step was still showing on staging and localhost causing some confusion.

Set default value to 0 so the step is disabled everywhere for the time being. Can be changed again in the future if we want to enable the user research question.


Davidslv (approved):
👍

PR #1145, opened 2021-01-27
Remove Brussels reference from PDF

zheileman:
Ticket: https://trello.com/c/xbDXxMyv

Just a small portion of copy removed as, after Brexit, Brussels 2 is no longer relevant to UK law.

Screenshot 2021-01-27 at 16 37 54


Davidslv (approved):
👍

PR #1146, opened 2021-01-28
Store the court county location code

zheileman:
Ticket: https://trello.com/c/IDdg4Kyo

As part of the centralisation work, we first need to take from CTF a piece of information we were not using before, and store it in our `Courts` DB table for later use.

In a follow-up PR we will print this code in the application PDF.


Davidslv (approved):
👍

PR #1152, opened 2021-02-23
Create list of centralised courts

zheileman:
Ticket: https://trello.com/c/ODIpiZet

This is a first step towards enabling centralisation, initially in just 5 courts, and then over the time more courts will be added to this list.

Currently this PR does not have any functionality other than create the list.

In follow-up PRs we will use this to know if a court is centralised or not and do different things based on that.


Davidslv (approved):
👍

PR #1164, opened 2021-03-11
Add centralisation metadata to the Audit table

zheileman:
As part of the centralisation rollout we are undergoing, some reports need to be generated to HMCTS so they know which completed applications are going to the central hub.

When there are few courts like now it can be done easily by court name, but soon there will be many more courts and generating these reports using the date and the court name is very tedious.

In order to simplify this we are adding to the metadata of any completed application a new attribute `centralised` true/false if the completed application has been for a centralised court.

This will allow to create reports much easily by just filtering by this flag.

A DB migration is needed to automate some already centralised courts in the first rollout.

### DIFF
--- a/app/lib/audit_helper.rb
+++ b/app/lib/audit_helper.rb
@@ -13,6 +13,7 @@ class AuditHelper
     {
       v: c100.version,
       postcode: postcode,
+      centralised: court.centralised?,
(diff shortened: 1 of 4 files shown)


Davidslv (approved):
👍

PR #1166, opened 2021-03-18
Replace Court Tribunal Finder endpoints with FaCT

Davidslv:
We have been requested to change the url to:

```
https://old.courttribunalfinder.service.gov.uk
```

**Their intention is to deploy this on Monday, 22 March 2021**.

A time has been requested so that we can try our best to sync a 
deployment,
this was not given at the time of this commit.

Test will fail until endpoint is live, visit: https://old.courttribunalfinder.service.gov.uk/search/results.json?aol=Children&postcode=MK93DX

Once live, re-run the tests.

Please see card: 
https://trello.com/c/WcVTq2VV/212-cas-court-tribunal-finder-url-change

---

⚠️ **Update 31/03/2021**

FaCT team has fixed all the showstoppers found with their API (refer to ticket https://trello.com/c/Pu2yNQXa for more details) and it seems there will be no need to use the `old` CTF subdomain as that was a workaround in case FaCT was released before fixing all the issues.

As a consequence, there have been a new commit in this PR (08b2607) to update the endpoints to call FaCT directly and we are awaiting a decision whether we are going to release this in the next few days.

### DIFF
--- a/app/views/about/contact.en.html.erb
+++ b/app/views/about/contact.en.html.erb
@@ -17,7 +17,7 @@
 
     <p class="govuk-body">
       We cannot answer any questions about applications that have already been submitted. You can
-      <a href="https://old.courttribunalfinder.service.gov.uk/search/spoe?aol=Children" class="govuk-link" rel="external" target="_blank">contact
+      <a href="https://www.find-court-tribunal.service.gov.uk/services/childcare-and-parenting/childcare-arrangements/search-by-postcode" class="govuk-link" rel="external" target="_blank">contact
         the relevant court</a> if you need to discuss your case.
     </p>
   </div>
--- a/app/views/about/contact.en.html.erb
+++ b/app/views/about/contact.en.html.erb
@@ -17,7 +17,7 @@
 
     <p class="govuk-body">
       We cannot answer any questions about applications that have already been submitted. You can
-      <a href="https://courttribunalfinder.service.gov.uk/search/spoe?aol=Children" class="govuk-link" rel="external" target="_blank">contact
+      <a href="https://old.courttribunalfinder.service.gov.uk/search/spoe?aol=Children" class="govuk-link" rel="external" target="_blank">contact
         the relevant court</a> if you need to discuss your case.
     </p>
   </div>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -836,7 +836,7 @@ en:
         edit:
           page_title: Details of urgent hearing
           heading: Details of urgent hearing
-          court_info_html: If you need to go to court now, <a href="https://old.courttribunalfinder.service.gov.uk/search/" rel="external" target="_blank" class="govuk-link">contact your nearest family court</a> for assistance.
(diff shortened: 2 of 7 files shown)


Davidslv:
[[CAS] Court Tribunal Finder URL Change](https://trello.com/c/WcVTq2VV/212-cas-court-tribunal-finder-url-change)

Davidslv:
For reference, we were initially asked to not verify SSL so I did this commit https://github.com/ministryofjustice/c100-application/pull/1166/commits/11bcf00b7660020660a98567064c0a5646d662e1

But in retrospective, In my opinion we should not disable SSL verification as this opens opportunities for [MITM attacks](https://en.wikipedia.org/wiki/Man-in-the-middle_attack) - even though they are hard, they are still possible.

I'd rather not take that risk in the middle of handover. So I've request their team to hold off this switch for a couple of days, and to get us a certificate.

app/views/about/contact.en.html.erb
```diff
@@ -17,7 +17,7 @@
 
     <p class="govuk-body">
       We cannot answer any questions about applications that have already been submitted. You can
-      <a href="https://old.courttribunalfinder.service.gov.uk/search/spoe?aol=Children" class="govuk-link" rel="external" target="_blank">contact
+      <a href="https://www.find-court-tribunal.service.gov.uk/services/childcare-and-parenting/childcare-arrangements/search-by-postcode" class="govuk-link" rel="external" target="_blank">contact
```
Davidslv, line 20: ✅

config/locales/en.yml
```diff
@@ -836,7 +836,7 @@ en:
         edit:
           page_title: Details of urgent hearing
           heading: Details of urgent hearing
-          court_info_html: If you need to go to court now, <a href="https://old.courttribunalfinder.service.gov.uk/search/" rel="external" target="_blank" class="govuk-link">contact your nearest family court</a> for assistance.
+          court_info_html: If you need to go to court now, <a href="https://www.find-court-tribunal.service.gov.uk/services/childcare-and-parenting/childcare-arrangements/search-by-postcode" rel="external" target="_blank" class="govuk-link">contact your nearest family court</a> for assistance.
```
Davidslv, line 839: :white_check_mark:

spec/mailers/notify_submission_mailer_spec.rb
```diff
@@ -123,7 +123,7 @@
         reference_code: '1970/01/4A362E1C',
         court_name: 'Test court',
         court_email: 'court@example.com',
-        court_url: 'https://old.courttribunalfinder.service.gov.uk/courts/test-court',
+        court_url: 'https://www.find-court-tribunal.service.gov.uk/courts/test-court',
```
Davidslv, line 126: :white_check_mark:

zheileman:
Let's hold on this, the certificate is not working yet, and also it may be we can go directly with FaCT (for the API) instead of the CTF workaround, as they are delaying the go-live and it will give them some extra time to fix the issues with the API so we may be able to use it from day one after all.

zheileman:
The `old` subdomain is now up and running and the SSL certificate is validated.

We will wait to merge this and release in the meantime, they will give us the go ahead, once FaCT has been released "officially".

zheileman (approved):
🥇

PR #1179, opened 2021-04-22
Revert "RST-3299 Remove the option for 'deal with another option not listed'"

garytaylor:
Reverts ministryofjustice/c100-application#1178

We have a query with this so need to stop it getting deployed for now until it is resolved


zheileman (approved):
👍

PR #1281, opened 2022-03-03
Add changes to the pipeline configuration to deploy in both the clusters

Ramparitala:
- Add changes to the pipeline configuration file to deploy in both the clusters.
- Removed  DEV block in the workflow as it is no longer needed.


zheileman (approved):
Looks good 👍 
We will know for sure once merged.

PR #1333, opened 2022-05-31
RST-4345 Fix duplicate child info when hitting back button

WillTaylor22:


app/controllers/concerns/names_crud_step.rb
```diff
@@ -22,6 +24,11 @@ def update
     )
   end
 
+  protected
+
+  def prevent_cache
+  end
```
zaparka: why is this method empty?

AdeyTaiwoSolirius (approved):

app/controllers/concerns/names_crud_step.rb
WillTaylor22: Good spot - was meant to be removed.

PR #1368, opened 2022-10-07
RST-4607

kalvinSolirius:
Added nil check to ensure code doesnt break


app/models/concerns/application_reference.rb
```diff
@@ -7,6 +7,7 @@ module ApplicationReference
   class_methods do
     def find_by_reference_code(ref)
       year, month, uuid_part = ref.split('/')
+      return if year.nil? && month.nil?
```
zaparka, line 10: Did you try || instead of &&. Does it pass or it could be just &&?

zaparka (approved):

PR #1395, opened 2022-11-11
RST-4755

kalvinSolirius:
Catch JSON errors and send email anyway.


app/presenters/summary/json_presenter.rb
```diff
@@ -49,5 +57,9 @@ def fee_amount
       amount = format("%.2f", (Rails.configuration.x.court_fee.amount_in_pence / 100))
       "£#{amount}"
     end
+
+    def this_will_error
```
zaparka, line 61: I think you can remove this

zaparka (approved):

PR #1556, opened 2024-04-02
RST-6302 Additional mediation exemptions

kalvinSolirius:


app/controllers/concerns/mediation_change.rb
```diff
@@ -0,0 +1,7 @@
+module MediationChange
+  extend ActiveSupport::Concern
+
+  def self.changes_apply?(c100_application)
+    c100_application.created_at >= DateTime.parse(ENV["MEDIATION_DATE"])
```
WillTaylor22, line 5: I'd suggest using Env.fetch and providing a safe default

WillTaylor22 (approved):

PR #1565, opened 2024-04-11
RST-6378 prevent application draft unless user has progressed

kalvinSolirius:


app/services/c100_app/save_application_for_later.rb
```diff
@@ -9,6 +9,8 @@ def initialize(c100_application, user)
     end
 
     def save
+      return false if c100_application.navigation_stack.size < 2
```
zaparka, line 12: I would have to look deeper at the context of this. But ideally you wouldn't check the size of the stack but position. But again I don't know enough about this :)

zaparka (approved):

zaparka (approved):

PR #1760, opened 2025-04-10
Minor gem updates

zaparka:
Technical debt


zaparka:
newer version present

